PR #27, opened 2018-02-15
* Enable claim free likecoin if not claimed

mckingho:


williamchong007:
what is Fix wrong email ..

PR #84, opened 2018-03-15
refactor to `/in` and `like.co`

williamchong007:


components/ReferralAction.vue
```diff
@@ -119,7 +119,7 @@ export default {
   },
   computed: {
     shareUrl() {
-      return `https://likecoin.store/register?from=${this.user}`;
+      return `https://like.co/register?from=${this.user}`;
```
nnkken: `/in`?

nuxt.config.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
   router: {
     extendRoutes(routes, resolve) {
       routes.push({
-        path: '/verify/:uuid/:coupon',
+        path: '/in/verify/:uuid/:coupon',
```
nnkken: `verify` should not be under `/in`?

PR #91, opened 2018-03-20
* Handle KYC steps and update styles

mileswong:


pages/in/tokensale/index.vue
```diff
@@ -70,7 +70,13 @@
                     <li>
                       <div>
                         <span class="label">{{ $t('TokenSale.label.token') }}</span>
-                        <span class="value highlight">LIKE</span>
+                        <a
+                          class="value highlight"
+                          href="https://etherscan.io/address/0x02F61Fd266DA6E8B102D4121f5CE7b992640CF98"
```
williamchong007: should use `LIKE_COIN_ADDRESS` from `@/constant/contract/likecoin/`

pages/in/tokensale/index.vue
```diff
@@ -70,7 +70,13 @@
...
+                          rel="noopener noreferrer"
```
williamchong007: probably no need `noreferrer` because we have `res.setHeader('Referrer-Policy', 'strict-origin');`

PR #110, opened 2018-03-22
= Sign plaintext instead of typedSignature for `user` and `kyc`

williamchong007:
Also does JSON beautify


williamchong007:
Also added the lost KYC unit test @Aludirk

PR #122, opened 2018-03-26
+Add ledger support for register and tokensale

williamchong007:
first 2 commit is https://github.com/likecoin/likecoin-store/pull/121


components/dialogs/MetamaskDialog.vue
```diff
@@ -58,6 +69,9 @@ export default {
       if (this.case === 'testnet') {
         return this.$t(`Dialog.metamask.content.switch${IS_TESTNET ? 'Rinkeby' : 'Main'}`);
       }
+      if (this.case === 'sign' && !this.isMetamask) {
+        return this.$t(`Dialog.metamask.content.${this.case}${this.webThreeType}`);
```
nnkken: `this.case === 'sign'`, so should use `'sign'` directly in the string?

williamchong007: 方便後面 copy and paste 或者改 case 唔會錯....

PR #147, opened 2018-04-03
Add redirect/email verify logic for `in-bundle`

williamchong007:


store/modules/actions/user.js
```diff
@@ -40,7 +41,7 @@ export async function getWalletByUser({ commit }, id) {
 export async function sendVerifyEmail({ commit, rootState }, { id, ref }) {
   /* guard ref to only 'tokensale' for now */
```
nwingt: Can remove this comment?

components/checkoutForm.vue
```diff
@@ -40,6 +40,7 @@
         </li>
       </ul>
     </div>
+    <email-dialog ref="emailDialog" :emailRef="'in-bundle'" />
```
nwingt: I think it is better to place the email dialog inside bundle page, as the child component should know nothing about its container

williamchong007: then the email checking logic (or even the checkout logic?) should be moved too?
or just fire event from inside and trigger `email-dialog` accordingly ?

nwingt: Maybe pass `emailRef` as props first

PR #254, opened 2018-04-16
Handle presale and bonus transaction records

mileswong:


components/TransactionHistory.vue
```diff
@@ -172,6 +191,8 @@ export default {
       return formatAmount(value, 'LIKE');
     },
     getFromToId(tx) {
+      if (this.isFromPresale(tx)) return 'presale';
+      if (this.isFromBonus(tx)) return 'bonus';
```
williamchong007: please change to `presalebonus`

PR #260, opened 2018-04-16
Fix #255

### LINKED ISSUE #255: [IE problems] above-the-fold problem with IE11
seems something invisible is blocking the whole above-the-fold ( `height 100%` ) for any cursor action
also vimeo player.js seems to have problem with ie11 (tested on win10, ie11, i5 haswell + R9 380X)

nwingt:
Fix IE not support fixed position inside a transformed element

### DIFF
--- a/assets/index.scss
+++ b/assets/index.scss
@@ -7,6 +7,7 @@
 }
 
 body {
+  position: static;
   z-index: -2;
 
   font-family: 'Open Sans',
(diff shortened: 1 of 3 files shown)


williamchong007:
有個 review tag 即係 expect merge 唔 merge 得

nwingt:
Wa

PR #270, opened 2018-04-18
+ Add getting started item

mileswong:


nwingt:
Please change `GettingStartedItem` to `TaskItem` more generic name

components/Mission/TaskItem.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <div :class="['task-item', state]" @click="$emit('click')">
+    <span class="lc-font-weight-600">{{ title }}</span>
```
nwingt: For repeatable component, please specify font-weight in css by using class `title-label` with `h1`
BTW Is there a TaskList component?

PR #280, opened 2018-04-19
+ Add Trust dialog UI

mileswong:


components/dialogs/TrustDialog.vue
```diff
@@ -0,0 +1,161 @@
+<template>
+  <base-dialog
+    :isShowDialog="isShowDialog"
```
nwingt: I don't know this is valid or not

williamchong007: probably not

PR #293, opened 2018-04-20
* Use material button in TrustDialog temporarily

mileswong:


components/dialogs/TrustDialog.vue
```diff
@@ -222,6 +224,8 @@ export default {
     text-align: center;
 
     .md-button.trust {
+      margin: 0;
+
       background-color: #3375bb;
```
nwingt: Please remove styles below

PR #297, opened 2018-04-23
Add bonus menu item

nwingt:


components/SlidingMenu/index.vue
```diff
@@ -25,7 +25,10 @@
                     :isHighlighted="true"
                     @click="onClickAccountButton" />
                 </li>
-                <li v-for="i in m.items" :key="i.key">
+                <li
+                  v-for="i in m.items"
+                  v-if="(i.isRegistered && getUserIsRegistered) || !i.isRegistered"
```
williamchong007: just `getUserIsRegistered || !i.isRegistered` will do?

nwingt: @Updated

PR #306, opened 2018-04-24
Add mission item placeholder

mileswong:


components/Mission/List.vue
```diff
@@ -23,6 +23,10 @@
             :is-referral="isReferral"
             @click="onClick(m)" />
         </li>
+
+        <li v-for="p in 3" :key="p.id" v-if="missions.length <= 0">
```
williamchong007: 3 name 成有意義嘅 var／constant 易睇d

PR #322, opened 2018-04-25
+ Add cumulative mission bonus history

williamchong007:
avoided need of passing `isHistory` by injecting `isHistory` into mission obj instead


pages/in/bonus/history.vue
```diff
@@ -70,29 +70,27 @@ export default {
...
+          this.refreshMissionHistoryList(this.getUserInfo.user);
         }
       }
     },
   },
   mounted() {
     if (!this.getUserIsFetching) {
       if (this.getUserIsRegistered) {
-        this.updateCompletedMission();
+        this.refreshMissionHistoryList();
```
nwingt: Need to pass `this.getUserInfo.user`?

williamchong007:
@nwingt updated

PR #332, opened 2018-04-26
Best effort mission sorting

williamchong007:
已知唔多 work，但係 QA(simonb) 會開心 d


store/modules/getters/mission.js
```diff
@@ -1,5 +1,12 @@
 import { MERGED_MISSIONS } from '@/constant';
 
+export const getProxyMissionReward = state => id => state.proxyBonus[id];
+
+const canCalim = (state, m) => (
```
nwingt: canClaim??

williamchong007:
@nwingt updated

PR #338, opened 2018-04-26
+ Add JoinTokenSale reward display hack

williamchong007:
since not better elegant way can be thought of

### DIFF
--- a/components/Mission/Item.vue
+++ b/components/Mission/Item.vue
@@ -84,9 +84,18 @@ export default {
           return `${new BigNumber(this.mission.pendingReferralBonus).div(ONE_LIKE).toFixed(2)} LIKE`;
         }
         return this.mission.referralReward;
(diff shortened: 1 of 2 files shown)


components/Mission/Item.vue
```diff
@@ -84,9 +84,20 @@ export default {
...
+        return `${this.getProxyMissionReward(this.mission.id)} LIKE`;
+      }
+      if (this.getUserInfo.referrer) {
+        /* joinTokenSale HACK */
+        if (this.mission.id === 'joinTokenSale' && this.mission.done) {
+          let reward = Number.parseInt(this.mission.reward, 10);
+          console.log(reward);
+          const extra = this.getProxyMissionReward('refereeTokenSale');
+          if (extra) reward += Number.parseInt(extra, 10);
+          console.log(extra);
```
nwingt: Can remove console.log?

PR #342, opened 2018-04-27
= Parse startTs in upcoming mission item

williamchong007:


components/dialogs/MissionDialog.vue
```diff
@@ -194,7 +195,11 @@ export default {
       const referralPostfix = this.isReferral ? 'Referral' : '';
       const { invitee } = this;
       return `
-        ${this.mission.upcoming ? this.$t('Mission.common.label.upcomingDate') : ''}
+        ${this.mission.upcoming ? this.$t(
+    'Mission.common.label.upcomingDate',
+    { upcoming: moment.utc(this.mission.upcoming).local().format('D-M-YYYY') },
+  )
+    : ''}
```
nwingt: The indentation is looking weird?

williamchong007: probably due to it is inside template string...
indent `properly` will case a linter error

PR #345, opened 2018-04-27
! Fix mission colors under different state

mileswong:


components/Mission/Item.vue
```diff
@@ -350,7 +361,7 @@ export default {
 .title-label,
 .reward-label {
   .upcoming & {
-    color: $like-gray-4;
+    color: $like-gray-4 !important;
   }
 }
```
nwingt: I suggest
```scss
&.upcoming  {
  .title-label,
  .reward-label {
    ...
  }
}
```

PR #347, opened 2018-04-27
Update mission list for mobile

nwingt:


components/Mission/InviteeGridList.vue
```diff
@@ -52,6 +40,9 @@ export default {
       type: Array,
       default: () => [],
     },
+    swiperId: {
+
+    },
```
williamchong007: `swiperId` 應該至少有 type？

components/Mission/InviteeGridList.vue
```diff
@@ -52,6 +40,9 @@ export default {
       type: Array,
       default: () => [],
     },
+    swiperId: {
+
```
mileswong: type: String ?

nwingt: Updated

components/Mission/InviteeGridList.vue
nwingt: Updated

components/Mission/List.vue
```diff
@@ -141,6 +178,11 @@ export default {
         }));
       }
     },
+    onClickSlide(i) {
+      if (this.missions.length > 0) {
+        this.onClick(this.missions[i]);
```
williamchong007: check 埋 `i < this.missions.length` ?

nwingt: Updated

PR #355, opened 2018-04-30
! Remove about LikeCoin menu item

nwingt:


williamchong007:
included in https://github.com/likecoin/likecoin-store/commit/d4177239ded1e72557b35b000156cfc66aab22f9

PR #357, opened 2018-04-30
Add mission completed dialog

nwingt:


components/dialogs/MissionDialog.vue
```diff
@@ -134,6 +136,11 @@
           </div>
           <!-- END - Invite Token Sale Section -->
         </div>
+        <mission-completed-banner
```
williamchong007: put `v-if="isCompleted"` in front, then `v-else` for main body seems more straight forward?

nwingt: @williamchong007 Updated

PR #361, opened 2018-04-30
* Swipe to mark invitee as seen at mobile

mileswong:


components/Mission/Swiper.vue
```diff
@@ -71,6 +71,11 @@ export default {
           el: this.getSelector('.swiper-pagination'),
           clickable: true,
         },
+        on: {
+          slideChange: () => {
+            this.onSlideChange();
+          },
+        },
```
nwingt: I prefer using `@slideChange`

PR #369, opened 2018-05-02
Use `window.location.assign` to open URL in Trust browser

nwingt:


components/dialogs/MissionDialog.vue
```diff
@@ -196,6 +196,7 @@ export default {
       'getPopupMission',
       'getMissionById',
       'getUserInfo',
+      'getWeb3Type',
```
williamchong007: 結果好似無用過？

nwingt: Updated

PR #375, opened 2018-05-03
+ Allow press articles to be opened automatically by query string

mileswong:


components/home/PressCoverage.vue
```diff
@@ -106,5 +121,23 @@ export default {
       presses,
     };
   },
+  mounted() {
+    const query = /[?&]press=([^&]*)/.exec(window.location.search);
```
nwingt: Can obtain `pressId` from `this.$route.query.press`

components/ImageGrid.vue
```diff
@@ -65,8 +65,11 @@ export default {
     };
   },
   methods: {
-    openLightBox(index) {
-      this.$refs.lightbox.showImage(this.imagesMap[index]);
+    openLightBox(id) {
+      const index = this.imagesMap[id];
+      if (index !== undefined && this.$refs.lightbox) {
```
williamchong007: `if (index &&` 簡單 d？

mileswong: it will be falsy if index is 0

PR #384, opened 2018-05-04
- Remove unused `InputDialog` and `showEmail` in UserInfoForm

williamchong007:


components/dialogs/InputDialog.vue
```diff
@@ -63,6 +63,12 @@ export default {
...
   },
   methods: {
+    show() {
+      this.$refs.dialog.show();
+    },
+    hide() {
+      this.$refs.dialog.hide();
+    },
     onCancel() {
       this.$refs.dialog.hide();
```
nwingt: `this.hide();`

PR #399, opened 2018-05-07
Fix number input for legacy iOS

nwingt:


nwingt:
@williamchong007 I tested `type="number"` is okay, but I don't have confidence in it, so I removed

PR #415, opened 2018-05-09
! Fix loggined user not auto redirected in ref pages

williamchong007:


pages/ref/_id/index.vue
```diff
@@ -129,6 +134,22 @@ export default {
...
+      }
+    },
+  },
+  mounted() {
+    if (!this.getUserIsFetching) {
+      if (this.getUserIsRegistered) {
+        this.$router.push({ name: 'in-tokensale' });
+      }
+    }
+  },
```
nwingt: Is it able to redirect to `in-tokensale` in `asyncData` func just like register page?

PR #453, opened 2018-05-15
= Stripe of dots in @gmail email instead

williamchong007:


server/api/users.js
```diff
@@ -173,7 +177,6 @@ router.put('/users/new', multer.single('avatar'), async (req, res) => {
       const referrerRef = await dbRef.doc(referrer).get();
       if (!referrerRef.exists) throw new Error('REFERRER_NOT_EXIST');
       if (!referrerRef.data().isEmailVerified) throw new Error('referrer email not verified');
-      if (referrerRef.data().isBlackListed) throw new Error('referrer limit exceeded');
```
nnkken: Why removing this?

server/api/users.js
```diff
@@ -186,7 +189,6 @@ router.put('/users/new', multer.single('avatar'), async (req, res) => {
       updateObj.email = email;
       updateObj.verificationUUID = FieldValue.delete();
       updateObj.isEmailVerified = false;
-      updateObj.lastVerifyTs = FieldValue.delete();
```
nnkken: Why removing this also?

williamchong007: becuz windows sourcetree sucks

PR #471, opened 2018-05-17
! Fix ico state is not reactive to time change

mileswong:


util/mixin/postICO.js
```diff
@@ -0,0 +1,28 @@
...
+      if (this.timer) {
+        clearInterval(this.timer);
+        this.timer = undefined;
+      }
+    },
+  },
+  mounted() {
+    if (!this.isICOEnded) {
+      this.timer = setInterval(() => {
+        this.isICOEnded = getIsTokenSaleEnded();
```
nnkken: should destroy timer after token sale ended?

PR #491, opened 2018-05-18
Fix token sale page not reactive when ICO ends

nwingt:


williamchong007:
@nwingt mark processing too if user has clicked the amount field? i.e. entering amount

pages/in/tokensale/index.vue
```diff
@@ -319,6 +320,9 @@ export default {
     pendingKYC() {
       return this.getUserInfo.pendingKYC || false;
     },
+    getIsICOEnded() {
+      return !this.isProcessingICO && this.getIsICOEnded;
```
mileswong: && `this.isICOEnded`?

PR #502, opened 2018-05-21
Add referrer dialog

nwingt:


components/LikeRegisterForm.vue
```diff
@@ -279,8 +292,30 @@ export default {
         }
       }
     },
+    async fetchReferrerInfo() {
+      this.shouldShowReferrerDialog = true;
+      this.setErrorDisabled(true);
+      try {
+        const response = await api.apiGetUserMinById(this.referrer);
```
williamchong007: probably want to use a action that call 
```
const { avatar, displayName } = await apiWrapper(ctx, api.apiGetUserMinById(r.id), { slient: true });
```
this kind of style is when apiWrapper didnt support `slient`

components/LikeRegisterForm.vue
```diff
@@ -279,8 +292,30 @@ export default {
         }
       }
     },
+    async fetchReferrerInfo() {
+      this.shouldShowReferrerDialog = true;
```
williamchong007: 未 load 到 referrer 嘅 data 嘅時候個 dialog 係咩樣的？

nwingt: Updated

components/LikeRegisterForm.vue
nwingt: Updated

PR #515, opened 2018-05-24
= Send server error as 500 instead of 400

williamchong007:


williamchong007:
@nwingt does express errorhandler handle exception properly, or do i have to call `next(err)` in each route?

nwingt:
@williamchong007 
You can just throw an error, express will automatically use the errorHandler
http://thecodebarbarian.com/80-20-guide-to-express-error-handling

williamchong007:
probably want to let the function decide how to print the error
moved the res part to error handler

williamchong007:
seems `throw err` won't trigger error handler, only `next(err)` do
updated

PR #527, opened 2018-06-01
+ Add send like and eth unit test

mckingho:
+ Add delegator account for unit test


test/stub/server/util/firebase.js
```diff
@@ -82,10 +89,12 @@ function collectionDoc(data, id) {
   let docObj;
   const obj = data.find(d => d.id === id);
   if (obj) {
+    // deep clone data object
+    const cloneObj = JSON.parse(JSON.stringify(obj));
```
nnkken: 1. quite adhoc and hacky...
2. may have performance issue, though should not be a problem
3. maybe extract into a function (e.g. `function deepCopy(obj) { ... }`, or use an existing library?

PR #528, opened 2018-06-04
* Update UI for register page and user info form

mileswong:


mileswong:
required email 要入返同 accept terms and policy 要tick，改緊

PR #535, opened 2018-06-06
+ Add prompt email notification dialog

mckingho:


components/dialogs/PromptNotificationDialog.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+  <base-dialog
+    ref="dialog"
```
nwingt: Please remove used ref

components/dialogs/PromptNotificationDialog.vue
```diff
@@ -0,0 +1,78 @@
...
+    ...mapActions([
+      'setPromptNotificationDialog',
+      'setNotification',
+    ]),
+    onEnableEmail() {
+      const {
+        user: id,
+      } = this.getUserInfo;
+      const user = this.getUserInfo;
+      this.setNotification({ id, isEmailEnabled: true, user });
```
nwingt: Not using `User.formatAndSignUserInfo`?

williamchong007: probably want to confirm UX with Kiu later

PR #539, opened 2018-06-07
! Fix unit test for register

mileswong:


util/test.js
```diff
@@ -0,0 +1,11 @@
+import { IS_TESTNET } from '@/constant';
+
+const getTestAttribute = name => (
+  IS_TESTNET ?
+    {
+      'lc-test': name,
+    } :
+    null
+);
```
nwingt: Wrap this into a function so that you can inject prefix, then directly use it for selector without #id 
```javascript
const getTestAttribute = prefix => (testId) => (
  IS_TESTNET ? ({ 
    'lc-test': `${prefix}-${testId}`,
  }) : (
    null
  )
);
```

components/LikeRegisterForm.vue
```diff
@@ -137,6 +137,8 @@ import MaterialButton from '~/components/MaterialButton';
 import { toDataUrl } from '@likecoin/ethereum-blockies';
 import { ETHERSCAN_HOST, W3C_EMAIL_REGEX, IS_TESTNET } from '@/constant';
 
+const getTestAttribute = getTestAttributePrefix('registerForm');
```
nwingt: Keep `getTestAttribute` for IDE look up and simplify into this
```javascript
methods: {
  ...
  getTestAttribute: getTestAttribute('registerForm'),
  ...
}
```

PR #547, opened 2018-06-11
* Add trade at QRYPTOS button

mileswong:


components/home/CTASection.vue
```diff
@@ -60,12 +60,16 @@
...
+                    <a
+                      :href="QRYPTOS_LIKEETH_URL"
+                      rel="noopener noreferrer"
+                      target="_blank">
+                      <material-button
+                        class="cta-btn support"
+                        :hasShadow="true">
+                        {{ $t('Home.Sale.button.tradeAtQRYPTOS') }}
+                      </material-button>
+                    </a>
```
williamchong007: not sure if good practice wrapping `<button>` with `<a>` ...

components/LikeCoinAmount.vue
```diff
@@ -17,14 +17,16 @@
...
+      <md-button
+        class="link what md-likecoin">
+        <a
+          class="lc-color-white"
+          :href="linkHref"
+          target="_blank"
+          rel="noopener noreferrer">
+          {{ linkText }}
+        </a>
+      </md-button>
```
williamchong007: wrapping `<a>` with `<button` on the other hand will cause confusion, when the button is clicked but <a> is not triggered
better use @click i suppose

nwingt: how about `a > button`, since we just need the style of the `<md-button />` and `<a />` will trigger the url tooltip at the bottom left corner when use hover it. We have lots of use cases `<nuxt-link />` wrapping `<md-button />`

williamchong007: just `a > button` then, but i suspect it somehow violate html spec, and `<nuxt-link />` wrappin `button` sounds like some quick hack left behind

nwingt: We found out that md-button has `href` props...

PR #554, opened 2018-06-13
Mission Url enhancement

mileswong:


pages/in/bonus/index.vue
```diff
@@ -291,6 +290,21 @@ export default {
         }
       }
     },
+    getMissionList(list) {
+      if (list.length > 0) {
+        if (this.$route.query.selectedMission) {
+          this.fetchSelectedMission({
```
williamchong007: probably should call this only if the selectedMission is not in the local list

server/api/mission.js
```diff
@@ -250,6 +250,65 @@ router.get('/mission/list/history/:id/bonus', jwtAuth, async (req, res, next) =>
...
+  try {
+    const { missionId, userId } = req.params;
+    const { userMissionList = [] } = req.query;
+
+    const [missionCol, userMission] = await Promise.all([
+      missionsRef.get(),
+      dbRef.doc(userId).collection('mission').doc(missionId).get(),
+    ]);
+
+    const getMission = id => missionCol.docs.find(d => d.id === id);
```
williamchong007: please add a comment stating that you need to trace back required mission chain, so need to fetch the whole doc instead of query

PR #561, opened 2018-06-15
! Hide subdescription from expired mission

mileswong:


components/dialogs/MissionDialog.vue
```diff
@@ -367,7 +367,8 @@ export default {
     },
     subDescription() {
       if (
-        !this.shouldShowDesktopOnly
+        !this.mission.isExpired
+        && !this.shouldShowDesktopOnly
         && this.hasReferrer
         && this.missionId === 'joinTokenSale'
```
williamchong007: should `isExpired` and `shouldShowDesktopOnly` outside of `joinTokenSale` related checking?

mileswong: Updated, and i think `shouldShowDesktopOnly` is not relevant to `joinTokenSale` as well as `joinTokenSale` is an expired mission so i removed it

PR #591, opened 2018-06-27
Add retweet mission

nwingt:
For BitMart campaign


components/forms/RetweetForm.vue
```diff
@@ -1,20 +1,17 @@
 <template>
-  <div class="retweet-form">
+  <div class="quote-tweet-form">
```
williamchong007: should it be `retweet-form` or is it intentional

nwingt:
@williamchong007 Updated error message handling

PR #612, opened 2018-07-09
Update transaction UI for high gas price

nwingt:


williamchong007:
@nwingt Please fix unit_test related cases

PR #620, opened 2018-07-12
! Fix inviteFriend mission always show completed banner after completed once

mileswong:
…ed once


components/dialogs/MissionDialog.vue
```diff
@@ -86,7 +86,7 @@
           </div>
 
           <mission-completed-banner
-            v-if="isCompleted || mission.isClaimed"
+            v-if="isCompleted || (mission.isClaimed && mission.id !== 'inviteFriend')"
```
williamchong007: mission 有支 flag 係 `staying`， 可能係 `ValidationHelper->filterMissionData()` 加返 `staying` send 俾 client？

mileswong: Updated

PR #633, opened 2018-07-16
! Fix web3 potential stuck due to no txHash

williamchong007:


server/util/web3.js
```diff
@@ -27,9 +27,13 @@ export function sendTransaction(tx) {
     const txEventEmitter = web3.eth.sendSignedTransaction(tx.rawTransaction);
     txEventEmitter.on('transactionHash', resolve)
       .on('error', (err) => {
-        if (err.message === 'Returned error: replacement transaction underpriced'
-          || err.message.includes('Returned error: known transaction:')) resolve(false);
-        else reject(err);
+        if (err.message === 'Returned error: replacement transaction underpriced') {
+          resolve(false);
+        } else (err.message.includes('Returned error: known transaction:')) {
+          resolve(web3.utils.sha3(tx.rawTransaction));
```
nnkken: can parse txHash from `err.message`?

williamchong007: not sure if trusting string from err msg or `web3.utils.sha3` is more unreliable..

PR #636, opened 2018-07-17
[WIP] Integrate payment flow

nwingt:


util/api/api.js
```diff
@@ -11,6 +11,8 @@ export const apiGetUserById = id => axios.get(`/api/users/id/${id}`);
 
 export const apiGetUserMinById = id => axios.get(`/api/users/id/${id}/min`);
 
+export const apiGetMerchantUserMinById = id => axios.get(`/api/users/merchant/${id}/min`);
```
williamchong007: Should it be called `apiGetUserMinByMerchantId` instead??

server/api/users.js
```diff
@@ -409,6 +409,23 @@ router.get('/users/id/:id/min', async (req, res, next) => {
...
+      const payload = query.docs[0].data();
+      if (!payload.avatar) payload.avatar = toDataUrl(payload.wallet);
+      payload.user = query.docs[0].id;
+      res.json(Validate.filterUserDataMin(payload));
+    } else {
+      res.sendStatus(404);
+    }
+  } catch (err) {
+    next(err);
+  }
```
williamchong007: please add simple unit test for this api

nwingt: Updated

PR #639, opened 2018-07-18
Refactor embed and button

nwingt:


williamchong007:
Does not pass stylelint, cannot deploy to rinkeby @nwingt

PR #645, opened 2018-07-20
Display LikeCoin account overview with USD amount in sliding menu

mileswong:


components/SlidingMenu/AccountOverview.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+  <div class="account-overview lc-margin-bottom-24-mobile">
```
nwingt: Please move `lc-margin-bottom-24-mobile` class in parent template, i.e.
```html
<account-overview class="lc-margin-bottom-24-mobile" ... />
```

test/api/misc.test.js
```diff
@@ -0,0 +1,12 @@
+import test from 'ava';
+import axios from 'axios';
+
+test('MISC: Get USD price from coingecko', async (t) => {
+  const res = await axios.get('https://api.coingecko.com/api/v3/coins/likecoin?localization=false')
+    .catch(err => err.response);
+
+  const { usd } = res.data.market_data.current_price;
```
nwingt: should wrap it in try catch

test/api/misc.test.js
```diff
@@ -0,0 +1,15 @@
...
+    const res = await axios.get('https://api.coingecko.com/api/v3/coins/likecoin?localization=false');
+
+    const { usd } = res.data.market_data.current_price;
+
+    t.is(res.status, 200);
+    t.true(usd > 0);
+  } catch (err) {
+    console.error(err); // eslint-disable-line no-console
+  }
+});
```
nwingt:
```javascript
let usd;
try {
  ...
  ({ usd } = res....)
} catch (err) {
  ...
}
t.true(usd !== undefined);
```

store/modules/ui.js
```diff
@@ -51,6 +52,7 @@ const state = {
   txDialogActionRoute: null,
   txDialogActionText: '',
   isShowingPromptNotificationDialog: false,
+  likeCoinUsdPrice: null,
```
williamchong007: is `likeCoinUsdPrice` a string? should it be `''` instead
also can the `action` / `getter` be better named to reflect it is a string, instead of number/bignumber?

PR #646, opened 2018-07-20
Fix payment flow

nwingt:


server/api/payment.js
```diff
@@ -133,6 +133,21 @@ router.post('/payment', async (req, res, next) => {
       LIKECOIN.LIKE_COIN_ADDRESS,
       txData,
     );
+    if (!toSubscriptionURL) {
+      res.json({ txHash });
+    }
```
williamchong007: sending both `res.json({ txHash });` here and below in 168 will cause error, on the other hand, why need to send early？

nwingt: Updated. This make sure sending res after `logTransferDelegatedTx` if user has `toSubscriptionURL`. If send earlier than that, oice cannot query the tx through tx API

williamchong007: just defer the `res.json({ txHash })` after `logTransferDelegatedTx` i guess

nwingt: Updated

PR #652, opened 2018-07-23
Show transfer amount on MetaMask Dialog on sign

mileswong:


mileswong:
Can be deployed if no other comments 👯‍♂️

PR #654, opened 2018-07-23
+ Add api for UI displaying liker list

williamchong007:
base on https://github.com/likecoin/likecoin-store/pull/653

!!! WANT COMMENT ON THE TITLE FETCHING CODE


pages/in/embed/_id/list.vue
```diff
@@ -0,0 +1,204 @@
...
+      if (match && match[1]) url = `https://medium.com/p/${match[1]}`;
+      promises.push(axios.get(url).catch(() => ''));
+    }
+    const [{ data: likees }, { data: totalData }, { data: html }] = await Promise.all(promises);
+    let title = '';
+    if (html) {
+      let match = html.match('<title>(.*?)</title>');
+      if (match && match[1]) [, title] = match;
+      match = html.match('"title":"(.*?)"');
+      if (match && match[1]) [, title] = match;
```
williamchong007: not sure how to handle this part better
some url returns `])}while(1);</x>` with a JSON like payload, not sure how to judge

pages/in/embed/_id/list.vue
```diff
@@ -0,0 +1,204 @@
...
+      const match = query.referrer.match(mediumRegEx);
+      if (match && match[1]) url = `https://medium.com/p/${match[1]}`;
+      promises.push(axios.get(url).catch(() => ''));
+    }
+    const [{ data: likees }, { data: totalData }, { data: html }] = await Promise.all(promises);
+    let title = '';
+    if (html) {
+      let match = html.match('<title>(.*?)</title>');
+      if (match && match[1]) [, title] = match;
+      match = html.match('"title":"(.*?)"');
```
nnkken: Won't work with titles like `"title":"\"quoted words\""`.
Also, escaped characters will be weird, since the `\` is not removed.
Need to really parse the JSON response instead of 'hacking' the content out from the response string.

pages/in/embed/_id/list.vue
```diff
@@ -0,0 +1,204 @@
...
+      apiGetLikeButtonTotalCount(params.id, query.referrer),
+    ];
+    if (query.referrer) {
+      let url = query.referrer;
+      const mediumRegEx = '^(?:https?:\\/\\/)?medium\\.com\\/media\\/[a-zA-Z0-9_]+\\?postId=([a-zA-Z0-9_]+)';
+      const match = query.referrer.match(mediumRegEx);
+      if (match && match[1]) url = `https://medium.com/p/${match[1]}`;
+      promises.push(axios.get(url).catch(() => ''));
+    }
+    const [{ data: likees }, { data: totalData }, { data: html }] = await Promise.all(promises);
```
nnkken: If promises.length < 3, the last pattern matching will throw error

pages/in/embed/_id/list.vue
nnkken: Since you are only interested in `obj.title`, which assumes that the JSON response is an object, how about stripping the content before the first `{` and after the last `}`?

pages/in/embed/_id/list.vue
nnkken: Maybe `const [{ data: likees }, { data: totalData }, { data: html } = {}] = await Promise.all(promises);`?
Seems making the code cryptic...

pages/in/embed/_id/list.vue
nnkken: BTW maybe `title = match[1]` is better for understanding

pages/in/embed/_id/list.vue
williamchong007: `= {}` seems good, and still not as cryptic as python...

pages/in/embed/_id/list.vue
williamchong007: but seems is this best effort extraction, I dont really know the structure of json
and parsing an arbitrary source of json sounds dangerous to me

pages/in/embed/_id/list.vue
williamchong007: linter like `[sth, title] = match` more than `title = match[1]` ..

PR #659, opened 2018-07-24
* Enhance likee list page UI

mileswong:


components/UserAvatar.vue
```diff
@@ -0,0 +1,112 @@
...
+      />
+    </transition>
+
+    <transition
+      name="lc-transition-default"
+      mode="out-in"
+    >
+      <nuxt-link
+        v-if="user.displayName"
+        :to="{ name: 'id', params: { id: user.id } }"
```
williamchong007: using `nuxt-link` in here will caused unwanted effect in `like-button` repo
i.e. redirecting to `like link` page
not sure whats best way to handle this though...

mileswong: confirm with kiu, open a new window would be good enough

williamchong007: 會唔會有 case 係，將來 like.co 其它位唔係咁用，想真係 nuxt link？
i.e. should it be configurable, or is this part of the purpose for this component now

mileswong: i think this component at this point is made for likecoin-button;
if like.co could use this in the future then it can be totally changed to nuxt-link?

PR #660, opened 2018-07-25
Complete like button UI

nwingt:


components/LikeButton.vue
```diff
@@ -0,0 +1,481 @@
...
+    onClickKnob(e) {
+      if (this.hasMovedKnob) return;
+
+      if (this.isLocalSuperLike) {
+        this.$emit('super-like', e, true);
+      } else {
+        this.isShowBubble = true;
+        setTimeout(() => {
+          this.isShowBubble = false;
+        }, 500);
```
williamchong007: should store timers in a data/data array, and `clearTimeout` on `beforeDestroy ` if possible

components/LikeButton.vue
```diff
@@ -0,0 +1,481 @@
...
+        this.isShowClapEffect = true;
+        this.$nextTick(() => {
+          this.isShowClapEffect = false;
+        });
+
+        this.$emit('like', e);
+      }
+    },
+    // eslint-disable-next-line func-names
+    onMovingKnob: _throttle(function (e) {
```
williamchong007: should prefer `requestAnimationFrame` instead of `_throttle`? not sure if there is actual logic requirement that needs `_throttle`

williamchong007: also maybe `_throttle` can be used as a guard when `requestAnimationFrame` does not exist, or maybe we just want to polyfill it, idk

nwingt: I don't think requestAnimationFrame is necessary in here as this function does not manipulate DOM directly for animation, instead it based on data changes and I think Vue will take care of that

williamchong007: but by limiting the `onMovingKnob` to 60ms throttle, the moving knob max fps is limited to `16.66`, regardless of device capability here?

PR #667, opened 2018-07-30
Add profile setting page

mileswong:


pages/in/settings/index.vue
```diff
@@ -350,8 +330,14 @@ export default {
         reader.readAsDataURL(files[0]);
       }
     },
-    onClickSocialMediaDisconnect(platform) {
+    async onClickSocialMediaDisconnect(platform) {
```
nwingt: why need async here?

constant/index.js
```diff
@@ -30,7 +30,7 @@ if (IS_TESTNET) whiteList.push('localhost');
 
 export const REDIRECT_WHITE_LIST = whiteList;
 
-export const REDIRECT_NAME_WHITE_LIST = ['in-tokensale', 'in-bundle'];
+export const REDIRECT_NAME_WHITE_LIST = ['in-tokensale', 'in-bundle', 'in-settings'];
```
williamchong007: `REDIRECT_NAME_WHITE_LIST` is only used in `sendVerifyEmail`, so if `in-settings` is not used in email then no need add this

components/SocialMediaConnect.vue
```diff
@@ -219,18 +219,16 @@ export default {
     },
     ...mapGetters(['getUserInfo']),
   },
+  watch: {
+    platforms: {
+      handler(p) {
+        this.updateSelectedFacebookPage(p);
```
williamchong007: since this function is explicitly named for facebook, the checking
```if (this.isLarge && platforms && platforms.facebook)``` should be done here, instead of inside the function?

williamchong007: ... or maybe rename this function to better reflect its usage? since I just realize it is also called in `mounted()`

constant/index.js
mileswong: User may change their email at settings page, would like to refer them back

components/SocialMediaConnect.vue
mileswong: any suggestion? `refreshSelectedFacebookPage`?

williamchong007: in `mounted()` and `watch` , sth like ```if (this.isLarge) this.refreshAddtionalDetails()```,
then inside `refreshAddtionalDetails()`
```if (this.platforms && this.platforms.facebook) 
 this.updateSelectedFacebookPage(this.platforms.facebook);```?

mileswong: Updated

williamchong007:
@mileswong conflict

mileswong:
@williamchong007 resolved

PR #669, opened 2018-07-31
Add oice mission

nwingt:


components/dialogs/MissionDialogContent/ConnectOice.vue
```diff
@@ -0,0 +1,178 @@
...
+          openURL(this, 'https://oice.com/profile', 'oice');
+          break;
+        case 2:
+          this.onComplete();
+          break;
+        default:
+      }
+      this.nextStep();
+    },
+    async onComplete({ isSlient }) {
```
mileswong: need to provide a default value `{}`, otherwise line 153 will have error

PR #679, opened 2018-08-02
! Fix coupon unit test

williamchong007:


williamchong007:
problem is on `develop` branch, rebase and reopening

PR #715, opened 2018-08-23
+ Add bonus cooldown UI

mileswong:


williamchong007:
加 clock，改 store，改 UI 可唔可以分三個 commit 囧

mckingho:
unit test 改 `.invitee-mission-list .username-label` 好似fail 了

components/Mission/Item.vue
```diff
@@ -27,7 +22,18 @@
             :mission-id="mission.id"
             class="mission-icon"
           />
-          <h1 class="title-label"><span>{{ title }}</span></h1>
+
+          <mission-countdown-timer
+            v-if="isCooldown"
+            :date="countdownTime"
+            :layout="layout"
+            :isFull="layout === 'small'"
```
nwingt: if `isFull` depends on `layout`, then no need to define props, use internal computed props instead

components/Mission/Item.vue
```diff
@@ -148,6 +185,29 @@ export default {
       if (this.mission.done) return 'completed';
       return (this.mission.status === 'pending') ? 'pending' : 'active';
     },
+    proxyBonusCountdown() {
+      if (this.mission.isProxy) {
+        const details = this.getProxyMissionDetails(this.mission.id);
+        if (this.mission.id === 'inviteFriend') {
+          if (details && details.earliestInvitee && details.earliestInvitee.bonusCooldown) {
```
nwingt: Better declare const just before use
```javascript
if (this.mission.isProxy) {
  if (this.mission.id === 'inviteFriend') {
    const details = ...
    if (details &&
       ...
```

PR #725, opened 2018-08-31
* Temporary hide QRYPTOS related call to action buttons

mileswong:


williamchong007:
add some `QRYPTOS ` specific comment near them, so that a `find all` can find them all out easily？

mileswong:
search QRYPTOS_LIKEETH_URL will find them all

PR #727, opened 2018-09-03
Add LikeButton settings page

mileswong:


util/ValidationHelper.js
```diff
@@ -135,9 +135,18 @@ export const ValidationHelper = {
...
   }) {
-    const data = { id: userId };
+    const data = {
+      displayName,
+      id: userId,
+      isPublic,
+      url,
+    };
     if (pages) data.pages = pages;
+    if (isPublic === undefined) data.isPublic = true;
```
williamchong007: setting default value `isPublic = true` when destructing in line 140 saves this line

server/api/social/index.js
```diff
@@ -97,6 +101,31 @@ router.post('/social/unlink/:platform', jwtAuth, async (req, res, next) => {
   }
 });
 
+router.post('/social/public', jwtAuth, async (req, res, next) => {
```
williamchong007: use `PATCH` would be more explicit

server/api/social/index.js
```diff
@@ -97,6 +101,31 @@ router.post('/social/unlink/:platform', jwtAuth, async (req, res, next) => {
...
+    const {
+      user,
+      platforms,
+    } = req.body;
+    if (req.user.user !== user) {
+      res.status(401).send('LOGIN_NEEDED');
+      return;
+    }
+
+    console.log(platforms, dbRef.batch);
```
williamchong007: useless `console.log`

server/api/social/index.js
```diff
@@ -27,8 +27,10 @@ router.get('/social/list/:id', async (req, res, next) => {
     const col = await dbRef.doc(username).collection('social').get();
     const replyObj = {};
     col.docs.forEach((d) => {
-      const { isLinked } = d.data();
-      if (isLinked) replyObj[d.id] = Validate.filterSocialPlatformPublic({ ...d.data() });
+      const { isLinked, isPublic } = d.data();
+      if (isLinked && isPublic !== false) {
```
williamchong007: why not `isPrivate` ? what is the default value for `isPublic`?
if it is default public, explicitly setting `isPrivate = true` seems better?

server/api/social/link.js
```diff
@@ -0,0 +1,114 @@
...
+      user,
+      link,
+    } = req.body;
+    if (req.user.user !== user) {
+      res.status(401).send('LOGIN_NEEDED');
+      return;
+    }
+
+    const {
+      id,
```
williamchong007: what happens when i hijack payload `id = facebook` ? will I be enable to make a trap facebook button in my profile?
Handling create id logic pure serverside seems better?

server/api/social/link.js
```diff
@@ -0,0 +1,114 @@
...
+          .where('order', '>=', order)
+          .where('order', '<', oldOrder)
+          .get();
+        orderOffset = 1;
+      } else if (oldOrder < order) {
+        query = await dbRef.doc(user).collection('social')
+          .where('order', '>', oldOrder)
+          .where('order', '<=', order)
+          .get();
+        orderOffset = -1;
```
williamchong007: @nnkken do we have a better way of doing this...

williamchong007: @mileswong even if we have to handle this `order` thingy like this
this have to be in a firestore transaction, otherwise it can race like hell
But I think we should figure out some better way to do ordering, instead adding it one by one...

components/settings/SocialMediaIcon.vue
```diff
@@ -0,0 +1,159 @@
...
+        'social-media-icon--hover': isHovered,
+      }
+    ]"
+  >
+    <lc-tooltip>
+      <div
+        slot="activator"
+        :class="['social-media-icon__icon-wrapper', {
+          'social-media-icon__icon-wrapper--link': platform.isExternalLink,
+        }]"
```
nwingt: Prefer this 
```vue
"[
  'social-media-icon__icon-wrapper',
  {
    'social-media-icon__icon-wrapper--link': platform.isExternalLink,
  },
]"
```

components/settings/ExternalLinksPanel.vue
```diff
@@ -0,0 +1,241 @@
...
+      this.hasNewLink = true;
+    },
+    spliceLoadingId(linkId) {
+      const index = this.loadingLinkIds.findIndex(id => id === linkId);
+      if (index >= 0) {
+        this.loadingLinkIds.splice(index, 1);
+      }
+    },
+    async onEnd({ newIndex, oldIndex }) {
+      // switch links order
```
nwingt: You mean swap?

server/api/social/link.js
```diff
@@ -74,6 +78,7 @@ router.put('/social/links/:linkId', jwtAuth, async (req, res, next) => {
     const promises = [];
 
     const updateObj = {};
+    if (iconType) updateObj.iconType = iconType;
```
williamchong007: should check if `iconType` is a valid iconType first

williamchong007:
also seems `vuedraggable` is missing in package-lock.json @mileswong

nuxt.config.js
```diff
@@ -96,6 +96,7 @@ module.exports = {
           'https://*.facebook.net',
           'https://checkout.stripe.com/',
           'https://*.doubleclick.net',
+          "'self'",
```
williamchong007: add
```
'https://rinkeby.like.co',
'https://like.co',
```
seems better?

mileswong:
Updated, will add test cases for social link later

mileswong:
@williamchong007 updated

constant/index.js
```diff
@@ -20,6 +20,9 @@ export const CONFIRMATION_NEEDED = 6;
 export const EMAIL_REGEX = '^[a-zA-Z0-9.!#$%&\'*/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$';
 export const W3C_EMAIL_REGEX = IS_TESTNET ? '.*' : EMAIL_REGEX;
 
+/* TEMP: reformat medium referrer into medium post */
+export const MEDIUM_REGEX = '^(?:https?:\\/\\/)?medium\\.com\\/media\\/[a-zA-Z0-9_]+\\?postId=([a-zA-Z0-9_]+)';
```
williamchong007: https://github.com/likecoin/likecoin-store/pull/728/files#diff-0391df65b1d58f7cc015bcd5ef24fb67R89
this one is more generic, since medium support subdomain

PR #742, opened 2018-09-12
New homepage UI

mileswong:


store/modules/actions/staticData.js
```diff
@@ -50,7 +50,7 @@ export async function fetchLikeSuggestionDetails({ commit, dispatch }, url) {
     commit(types.STATIC_DATA_SET_LIKE_SUGGEST_DETAIL, { url, info });
     return { info };
   } catch (err) {
-    console.error(err);
+    commit(types.STATIC_DATA_SET_LIKE_SUGGEST_DETAIL, { url });
```
williamchong007: if the purpose of this is to remove entry, explicitly pass `null` as 2nd param seems more readable

PR #749, opened 2018-09-17
* Update homepage UI

mileswong:


components/LikerList.vue
```diff
@@ -0,0 +1,168 @@
...
+    likers() {
+      return this.getLikerListDetails(this.referrer).likers || [];
+    },
+    numOfLikes() {
+      return this.getLikerListDetails(this.referrer).numOfLikes || 0;
+    },
+  },
+  watch: {
+    likers(list) {
+      this.isShowAll = list.length <= 8;
```
williamchong007: please name `8` as a `const`

components/LikerList.vue
```diff
@@ -0,0 +1,168 @@
...
+    },
+  },
+  mounted() {
+    const { id, referrer } = this;
+    this.fetchLikeButtonStatistics({ id, referrer });
```
williamchong007: since `getLikerListDetails` is already in store, is there anyway to prevent repeated fetching?
or is repeated fetching intentional as a update mechanism?

PR #756, opened 2018-09-19
Homepage UI enhancements

mileswong:


components/home/OdometerCounter.vue
```diff
@@ -0,0 +1,111 @@
+<template>
+  <div class="odometer-counter">
+    <template v-if="numStr !== '0'">
```
williamchong007: would it be possible that `numStr` is falsy or negative?

mileswong: updated

PR #763, opened 2018-09-21
! Fix email not working social links

mileswong:


components/SocialMediaConnect.vue
```diff
@@ -294,9 +294,9 @@ export default {
       if (!isConnected) {
         this.connect(socialMedia);
       } else {
-        const url = this.getSocialMediaUrl(socialMedia);
+        const { url } = platform;
```
williamchong007: where is `getSocialMediaUrl` used now? what is it actual function is not used now

mileswong: updated

components/SocialMediaConnect.vue
```diff
@@ -294,12 +294,14 @@ export default {
...
         if (url) {
-          const isEmail = !IS_TESTNET && new RegExp(W3C_EMAIL_REGEX).test(url);
+          const isEmail = new RegExp(EMAIL_REGEX).test(url);
           let urlPath = url;
```
williamchong007:
```
let urlPath = this.getSocialMediaUrl(socialMedia);
if (isEmail) {
  urlPath = `mailto:${url}`;
}
```
seems better?

PR #764, opened 2018-09-21
* Upgrade to node:10

williamchong007:


williamchong007:
@nnkken end up i need `unsafe-perm` to run `npm postinstall`

PR #766, opened 2018-09-21
[WIP] JWT revamp

williamchong007:
Done:
+ Add AUD ISS and permissions field
+ Add RSA support
+ Add jwtId for session control

TODO in future:
+ Add void of session

### DIFF
--- a/server/api/users.js
+++ b/server/api/users.js
@@ -14,7 +14,7 @@ import axios from '../../plugins/axios';
 import { ValidationHelper as Validate, ValidationError } from '../../util/ValidationHelper';
 import { personalEcRecover, web3 } from '../util/web3';
 import { uploadFileAndGetLink } from '../util/fileupload';
-import { jwtSign, jwtAuth, jwtVerify } from '../util/jwt';
+import { jwtSign, jwtAuth } from '../util/jwt';
 import publisher from '../util/gcloudPub';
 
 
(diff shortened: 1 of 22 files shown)


williamchong007:
`setSessionCookie()` seems very questionable now, maybe need to discuss

server/api/users.js
```diff
@@ -81,15 +81,21 @@ function setSessionCookie(req, res, token) {
...
     wallet,
     permissions: ['read', 'write', 'like'],
   };
-  const token = jwtSign(payload);
+  const { token, jwtid } = jwtSign(payload);
   res.cookie('likecoin_auth', token, AUTH_COOKIE_OPTION);
   setSessionCookie(req, res, token);
+  await dbRef.doc(user).collection('session').doc(jwtid).create({
+    lastAccessedUserAgent: req.headers['user-agent'] || 'unknown',
+    lastAccessedIP: req ? (req.headers['x-real-ip'] || req.ip) : undefined,
```
nnkken: So it is possible to have falsy value `req`? Then `setSessionCookie` should care about this?

server/api/users.js
```diff
@@ -320,14 +326,19 @@ router.put('/users/new', apiLimiter, multer.single('avatar'), async (req, res, n
...
   const { wallet } = req.body;
   if (req.user.wallet !== wallet) {
     res.status(401).send('LOGIN_NEEDED');
     return;
   }
   setSessionCookie(req, res, req.cookies.likecoin_auth);
   res.sendStatus(200);
+  await dbRef.doc(req.user.user).collection('session').doc(req.user.jti).update({
+    lastAccessedUserAgent: req.headers['user-agent'] || 'unknown',
+    lastAccessedIP: req ? (req.headers['x-real-ip'] || req.ip) : undefined,
```
nnkken: Same as line 95

server/api/users.js
williamchong007: actually not possible here, copy and paste programming fault..

PR #788, opened 2018-10-15
+ Add wallet sign in and sign up

mckingho:


components/EthMixin.vue
```diff
@@ -0,0 +1,44 @@
...
+      this.wallet = this.getLocalWallet;
+      if (this.getIsWeb3Polling) {
+        this.stopWeb3Polling();
+        this.callbackFunc();
+      }
+    },
+  },
+  methods: {
+    ...mapActions([
+      'refreshUser',
```
nwingt: Please remove if not use in here

PR #793, opened 2018-10-23
Update AuthDialog

nwingt:


server/api/users.js
```diff
@@ -311,29 +353,57 @@ router.post('/users/login', async (req, res, next) => {
...
-        user = query.docs[0].id;
+
+      case 'facebook': {
+        try {
+          const { accessToken, platformUserId } = req.body;
+          await fetchFacebookUser(accessToken);
+          const query = (
+            await oAuthDbRef
+              .where(`${platform}UserId`, '==', platformUserId)
+              .get()
```
williamchong007: probably want to add `.limit(1)` for query that only checks for existence
```suggestion
              .get().limit(1)
```
(although this part was originally wrote by me and idk why I didn't add `limit()`..)

components/dialogs/AuthDialog.vue
```diff
@@ -318,25 +330,110 @@ export default {
...
+                    userID: platformUserId,
+                  } = authResponse;
+
+                  // Get user info
+                  window.FB.api(
+                    '/me?fields=name,email',
+                    ({
+                      email,
+                      name,
+                    }) => {
```
williamchong007: user info should be get only for register prefill

server/api/users.js
```diff
@@ -61,26 +63,49 @@ const apiLimiter = new RateLimit({
...
+      case 'twitter': {
+        const { firebaseIdToken } = req.body;
+        ({ uid: firebaseUserId } = await admin.auth().verifyIdToken(firebaseIdToken));
+        payload = req.body;
+        break;
+      }
+
+      case 'facebook': {
+        const { accessToken } = req.body;
+        await fetchFacebookUser(accessToken);
```
williamchong007: should get verified info from `fetchFacebookUser()` and match/replace against untrusted payload from client

server/api/users.js
```diff
@@ -91,14 +116,13 @@ router.post('/users/new', apiLimiter, multer.single('avatar'), async (req, res,
       referrer,
       locale = 'en',
       accessToken,
+      platformUserId,
```
williamchong007: should not trust `platformUserId` directly from client, see above

server/api/users.js
```diff
@@ -91,14 +116,13 @@ router.post('/users/new', apiLimiter, multer.single('avatar'), async (req, res,
...
       secret,
+      isEmailVerified,
```
williamchong007: should not trust `isEmailVerified` directly from user, should only trust when 
1. oauth provider provides a trusted email
2. oauth provider email == user input email

server/api/users.js
```diff
@@ -163,6 +189,11 @@ router.post('/users/new', apiLimiter, multer.single('avatar'), async (req, res,
       locale,
     };
 
+    if (email) {
+      createObj.email = email;
+      createObj.isEmailVerified = getBool(isEmailVerified);
```
williamchong007: should write separated logic for `isEmailVerified`, see above

server/api/users.js
```diff
@@ -182,6 +213,17 @@ router.post('/users/new', apiLimiter, multer.single('avatar'), async (req, res,
       await dbRef.doc(referrer).collection('referrals').doc(user).create(timestampObj);
     }
 
+    switch (platform) {
+      case 'facebook': {
+        const doc = {
+          [`${platform}UserId`]: platformUserId,
```
williamchong007: maybe we can use an object, sth like `${platform}.userId` which (should be) also query-able in firestore
it depends on how we going to get `social/links` though

server/api/users.js
```diff
@@ -311,29 +353,57 @@ router.post('/users/login', async (req, res, next) => {
...
-      const { firebaseIdToken } = req.body;
-      const { uid: firebaseUserId } = await admin.auth().verifyIdToken(firebaseIdToken);
-      const query = await dbRef.where('firebaseUserId', '==', firebaseUserId).get();
-      if (query.docs.length > 0) {
-        user = query.docs[0].id;
+
+      case 'facebook': {
+        try {
+          const { accessToken, platformUserId } = req.body;
+          await fetchFacebookUser(accessToken);
```
williamchong007: should verify user payload agains `fetchFacebookUser()`, see above

util/FirebaseApp.js
```diff
@@ -43,9 +43,25 @@ export function getFirebaseProvider(platform) {
...
+    emailVerified: isEmailVerified = false,
+    photoURL: avatarURL = false,
+  } = result.user;
   const { accessToken, secret } = result.credential;
   const firebaseIdToken = await firebase.auth().currentUser.getIdToken();
-  return { accessToken, secret, firebaseIdToken };
+  return {
+    displayName,
+    email,
+    isEmailVerified,
```
williamchong007: probably dont want to expose `isEmailVerified` for client code

server/util/api/users.js
```diff
@@ -113,6 +116,8 @@ async function userInfoQuery({
   wallet,
   email,
   firebaseUserId,
+  platform,
+  platformUserId,
```
williamchong007: changes on `userInfoQuery()` should be reflected in `checkUserInfoUniqueness()` and `checkIsOldUser()`

components/dialogs/AuthDialog.vue
```diff
@@ -318,25 +330,110 @@ export default {
...
+        case 'google':
+        case 'twitter':
           this.signInPayload = await firebasePlatformSignIn(platform);
-          this.login();
-        }
+          break;
+
+        case 'facebook':
+          try {
+            this.currentTab = 'loading';
```
williamchong007: make it a method `setCurrentTabLoading()` sounds more robust?

nwingt: Probably change it later

util/FirebaseApp.js
nwingt: Registration form depends on it for toggling email field

components/dialogs/AuthDialog.vue
```diff
@@ -318,25 +330,86 @@ export default {
...
     },
+    async signInWithWallet() {
+      this.currentTab = 'signingIn';
+
+      // Determine whether the wallet has registered
+      try {
+        await apiCheckIsUser(this.getLocalWallet);
+      } catch (err) {
+        // Assume it is 404
+        // Redirect user to register
```
williamchong007: actually can check if it is actually 404...

components/dialogs/AuthDialog.vue
```diff
@@ -318,25 +330,86 @@ export default {
...
+        };
+        this.currentTab = 'register';
+        return;
+      }
+
+      try {
+        this.signInPayload = await User.signLogin(this.getLocalWallet);
+        this.login();
+      } catch (err) {
+        if (err.message.indexOf('Invalid "from" address') >= 0) {
```
williamchong007: TODO: should be guarded by actually retrieving the wallet address before asking user to sign

components/dialogs/AuthDialog.vue
```diff
@@ -318,25 +330,86 @@ export default {
...
+          this.currentTab = 'loading';
+          this.startWeb3AndCb(this.signInWithWallet, true);
+        } else {
+          // Return to login portal if user denied signing
+          this.currentTab = 'portal';
```
williamchong007: TODO: should actually check error message for user reject signing string

components/dialogs/AuthDialogContent/Register.vue
```diff
@@ -131,6 +131,7 @@
 </template>
 
 <script>
+import axios from 'axios';
```
williamchong007: Does `import axios from '~/plugins/axios'` also work? if yes should prefer that instance

nwingt: that instance has predefined baseURL point to server which is not the use case

nwingt: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.

components/dialogs/AuthDialog.vue
```diff
@@ -388,8 +388,11 @@ export default {
       try {
         await apiCheckIsUser(this.getLocalWallet);
       } catch (err) {
-        // Assume it is 404
-        // Redirect user to register
+        if (!err.response || err.response.status !== 404) {
+          this.currentTab = 'signInError';
+          return;
```
williamchong007: prefer `if (status === 404) ` so that later it can be easily swapped to a `switch` if needed?

components/dialogs/AuthDialog.vue
```diff
@@ -421,9 +427,12 @@ export default {
         this.setUserNeedAuth(false);
         this.redirectToUserPage();
       } catch (err) {
-        console.error(err);
-        // TODO: Check error
-        // Assume it is 404
+        if (!err.response || err.response.status !== 404) {
+          this.currentTab = 'signInError';
+          return;
+        }
```
williamchong007: same as above, `404` should be explicitly handle as a case

components/dialogs/AuthDialog.vue
nwingt: Updated

components/dialogs/AuthDialog.vue
nwingt: Updated

components/dialogs/AuthDialog.vue
williamchong007: @mckingho please look into why we are not polling address change later

PR #798, opened 2018-10-30
Enable OAuth login linking

nwingt:


server/util/api/users.js
```diff
@@ -232,3 +232,62 @@ export async function checkReferrerExists(referrer) {
...
+  likeCoinId,
+  platform,
+}) {
+  // Check if OAuth doc exists
+  const oAuthDocRef = oAuthDbRef.doc(likeCoinId);
+  const oAuthDoc = await oAuthDocRef.get();
+  if (!oAuthDoc.exists) return;
+
+  const data = oAuthDoc.data();
+  const isSole = Object.keys(data).length < 2;
```
williamchong007: prefer `<= 1` for more intuitive reading?

server/util/api/users.js
```diff
@@ -232,3 +232,62 @@ export async function checkReferrerExists(referrer) {
...
+      wallet,
+    } = userDoc.data();
+    if (firebaseUserId || wallet) {
+      await oAuthDocRef.delete();
+    } else {
+      throw new ValidationError('USER_UNLINK_SOLE_OAUTH_LOGIN');
+    }
+  } else {
+    delete data[platform];
+    await oAuthDocRef.set(data);
```
williamchong007: can use object update with delete value instead?
https://firebase.google.com/docs/firestore/manage-data/add-data#update_fields_in_nested_objects

components/dialogs/AuthDialog.vue
```diff
@@ -377,10 +383,21 @@ export default {
...
       this.currentTab = 'loading';
-      await firebaseSendSignInEmail(email);
-      this.currentTab = 'checkInbox';
+
+      if (this.isSigningInWithEmail) {
+        try {
+          this.signInPayload = await firebaseHandleSignInEmailLink(email);
+          this.login();
+        } catch (err) {
+          this.currentTab = 'signInError';
```
williamchong007: will we log the `err` somehow? at least to sentry

util/FirebaseApp.js
```diff
@@ -97,31 +97,30 @@ export function firebaseIsSignInEmailLink() {
...
-export async function firebaseHandleSignInEmailLink() {
+export async function firebaseHandleSignInEmailLink(providedEmail) {
   let email;
-  try {
-    ({ email } = JSON.parse(window.localStorage.getItem('emailForSignIn')));
-  } catch (err) {
-    return null;
+
+  if (providedEmail) {
+    email = providedEmail;
```
williamchong007: in what case the `email` will be `provided`?

server/api/users.js
```diff
@@ -225,17 +239,13 @@ router.post('/users/new', apiLimiter, multer.single('avatarFile'), async (req, r
...
-        await oAuthDbRef.doc(user).create(doc);
-        break;
-      }
-      default:
+    if (platformUserId) {
+      const doc = {
+        [platform]: {
+          userId: platformUserId,
+        },
+      };
```
williamchong007: probably want to add a comment that, this `platform` is guarded from user modification by the fact that only `platformUserId` is only set when the `platform` is valid.
or even better, have a function that filters valid platforms in this case? not sure

server/api/users.js
```diff
@@ -222,6 +222,14 @@ router.post('/users/new', apiLimiter, multer.single('avatarFile'), async (req, r
     if (email) {
       createObj.email = email;
       createObj.isEmailVerified = isEmailVerified;
+
+      if (isEmailVerified) {
+        await dbRef
+          .doc(user)
+          .collection('mission')
+          .doc('verifyEmail')
+          .set({ done: true }, { merge: true });
```
williamchong007: please state with comment this is a hack, since we should rely on firestore trigger to mark the mission done, which in reality is too slow for new user UX...

williamchong007:
@nwingt this PR is getting huge, maybe you want to merge the part that was reviewed first?

components/dialogs/WalletNoticeDialog.vue
```diff
@@ -0,0 +1,477 @@
...
+          ref="page1"
+          :key="1"
+          class="wallet-notice-dialog__content"
+        >
+          <p class="wallet-notice-dialog__content__description">
+            {{ $t('WalletNoticeDialog.page[0].description') }}
+          </p>
+          <ul class="wallet-notice-dialog__content__list">
+            <li
+              v-for="i in 3"
```
williamchong007: is `3` `PAGE_COUNT` or something else？

components/dialogs/WalletNoticeDialog.vue
```diff
@@ -0,0 +1,477 @@
...
+          height: `${elem.offsetHeight}px`,
+        };
+      }
+    },
+    onCancel() {
+      this.updateIsShow(false);
+      this.$emit('cancel');
+    },
+    onProceed() {
+      if (this.pageIndex < 2) {
```
williamchong007: is `2` related to `PAGE_COUNT - 1`?

components/dialogs/WalletNoticeDialog.vue
nwingt: 3 is number of bullet points in page 1

nwingt: comment added

PR #805, opened 2018-11-05
Strip away subscription & Stripe & merge master

nwingt:


williamchong007:
@nwingt might want to leave the `GET` and some frontend logic behind for displaying manual subscription, but need to talk with kiu and @mckingho for details

PR #819, opened 2018-11-14
+ Add referral data in registration log

mckingho:


components/dialogs/AuthDialog.vue
```diff
@@ -327,6 +330,9 @@ export default {
         }
       }
     }
+
+    this.referrer = this.$route.query.from;
+    this.sourceURL = this.$route.query.referrer;
```
nnkken: Seems interchanged, please confirm

nwingt: Confirmed

PR #826, opened 2018-11-15
! Fix misc UI bug and add disable editing email when sign up

mckingho:


pages/in/index.vue
```diff
@@ -367,6 +371,8 @@ export default {
   &__cta {
     flex: 1;
 
+    margin-top: 16px;
```
nwingt: Can use `.lc-margin-top-16` class

components/dialogs/AuthDialog.vue
```diff
@@ -276,8 +276,9 @@ export default {
     shouldShowDialog() {
       return this.getIsShowAuthDialog && !this.shouldHideDialog;
     },
-    shouldSignUpShowEmail() {
-      return !(this.signInPayload.isEmailVerified && this.platform === 'google');
+    shouldSignUpEditEmail() {
```
nwingt: `isEmailEditable` is more concise

components/dialogs/AuthDialog.vue
```diff
@@ -276,8 +276,9 @@ export default {
...
+      return (!(this.signInPayload.isEmailVerified && this.platform === 'google')
+        && this.platform !== 'email');
```
nwingt: More readable
```js
return (
   !(this.signInPayload.isEmailVerified && this.platform === 'google')
   && this.platform !== 'email'
);

PR #828, opened 2018-11-15
* Display different text when user need to re-enter email

nwingt:


components/dialogs/AuthDialog.vue
```diff
@@ -74,6 +74,7 @@
         >
           <email-signin-form
             :is-show-back="!isSigningInWithEmail"
+            :is-re-enter="errorCode = 'FIREBASE_EMAIL_LINK_AUTH_NO_EMAIL'"
```
williamchong007: `errorCode === 'FIREBASE_EMAIL_LINK_AUTH_NO_EMAIL'` ？

PR #829, opened 2018-11-15
= Use BigNumber 0 for more robust behaviour in likecoin amount

williamchong007:


server/api/users.js
```diff
@@ -441,6 +441,10 @@ router.post('/users/login', async (req, res, next) => {
                 const userQuery = await dbRef.where('email', '==', email).get();
                 if (userQuery.docs.length > 0) {
                   const [userDoc] = userQuery.docs;
+                  const { firebaseUserId: userFirebaseId } = userDoc.data();
```
nnkken: The `firebaseUserId ` from `userDoc` is mapped into `userFirebaseId`, which is then compared with `firebaseUserId`, which is the `uid` from firebase...
The names are too confusing, please consider better names

williamchong007: reopening in different PR

PR #848, opened 2018-11-19
Fix social media issues

nwingt:


williamchong007:
`isHideLegacy` is better? We want to hide it in only few occasions

PR #849, opened 2018-11-19
+ Add register GA, login, logout log

mckingho:


server/api/users.js
```diff
@@ -523,9 +545,37 @@ router.post('/users/login', async (req, res, next) => {
...
 
-router.post('/users/logout', (req, res) => {
-  clearAuthCookies(req, res);
-  res.sendStatus(200);
+router.post('/users/logout', jwtAuth('read'), async (req, res, next) => {
+  try {
+    clearAuthCookies(req, res);
+    res.sendStatus(200);
+
+    if (req.user.user) {
```
williamchong007: does `clearAuthCookies()` affect value of  `req.user.user` ?
If yes you might want to store it in a var first, if not just ignore me

components/dialogs/AuthDialog.vue
```diff
@@ -291,6 +293,11 @@ export default {
         }
       }
     },
+    currentTab(tab) {
+      if (tab === 'register') {
+        logTrackerEvent(this, 'RegFlow', 'ShowRegisterForm', 'ShowRegisterForm', 1);
```
williamchong007: to prevent fire multiple times, add a local variable to store if you have fire this event before?
you clear it on `mounted` or `created` if needed

PR #850, opened 2018-11-19
 [WIP] + Add temp UI for oauth connect

williamchong007:


store/modules/actions/user.js
```diff
@@ -61,12 +62,33 @@ export async function loginUser({ commit, dispatch }, data) {
   return true;
 }
 
-export async function linkWalletToUser({ commit, dispatch }, data) {
-  await apiWrapper({ commit, dispatch }, api.apiLinkAuthWallet(data), { blocking: true });
+export async function linkWalletToUser({ commit, dispatch }, payload) {
+  await apiWrapper({ commit, dispatch }, api.apiLinkAuthPlatform('wallet', payload), { blocking: true });
   await dispatch('refreshUser');
   return true;
 }
```
nwingt: Just use linkUserAuthPlatform directly for linking wallet ?

williamchong007: wallet cannot be unlinked, might want to give it special privilege of a special naming for that

PR #874, opened 2018-11-29
Update settings page UI

nwingt:


williamchong007:
does this unify twitter social link to firebase also?

PR #876, opened 2018-11-30
Fix unit test

nwingt:


test/data/tx.json
```diff
@@ -3,8 +3,8 @@
         {
             "id": "0x2d9549ba2d895141660d8fb92bc63ce7caf9eae104550d1ac77a6cd82daf0cad",
             "txHash": "0x2d9549ba2d895141660d8fb92bc63ce7caf9eae104550d1ac77a6cd82daf0cad",
-            "from": "0x98e6269f33d3191b6e6DEdD01e93c51cc14257d3",
-            "to": "0x83D3F8effbf3924D34F51531ce57C97C9fA2E5CF",
+            "from": "0x4b25758E41f9240C8EB8831cEc7F1a02686387fa",
+            "to": "0x76ECdE1CC1D2D87D7f20c2C6EAe7529a7379920c",
```
williamchong007: why are these changed?

nwingt: This fixes 401 error as no user is found by the JWT token from the address

store/index.js
```diff
@@ -10,10 +10,12 @@ import eth from './modules/eth';
...
 const createStore = (() => new Vuex.Store({
   actions: {
-    async nuxtServerInit({ commit }, { req }) {
+    async nuxtServerInit({ commit }, { req, route }) {
       /* TODO: actually try to verify jwt first? */
       commit('USER_SET_USER_INFO', {});
-      commit('USER_SET_AFTER_AUTH_ROUTE', null);
+      if (route.name !== 'in-register') {
+        commit('USER_SET_AFTER_AUTH_ROUTE', null);
+      }
```
williamchong007: why check for register page here?

plugins/testing.js
```diff
@@ -0,0 +1,4 @@
+import Vue from 'vue';
+import { getSingleTestAttribute } from '../util/test';
+
+Vue.prototype.$testID = getSingleTestAttribute;
```
williamchong007: What does this do? can this be excluded when production build?

store/index.js
nwingt: the redirection call will execute `nuxtServerInit`, which clear the `preAuthRoute`, before entering register page

The issue
1. Enter auth required page e.g. `/in/bonus`
2. Set preAuthRoute
3. Redirect to register page
4. Reset preAuthRoute 💥

PR #887, opened 2018-12-05
* Update user data to intercom

mckingho:


components/footer/Footer.vue
```diff
@@ -69,6 +77,8 @@ export default {
           const amount = await EthHelper.queryEthBalance(wallet);
           opt.ETH = Number(new BigNumber(amount).dividedBy(ONE_LIKE).toFixed(4));
         }
+        opt.isEmailVerified = !!isEmailVerified;
+        opt.isAvatarUpdated = this.AVATAR_DEFAULT_PATH !== avatar;
```
nwingt: this boolean is more likely `hasAvatar`

mckingho: This part is moved to cloud function, so it is updated at firebase repo.

components/footer/Footer.vue
```diff
@@ -91,6 +92,18 @@ export default {
         this.$intercom.trackEvent('likecoin-store_error', { message });
       }
     },
+    getUserSocialPlatforms(platforms) {
+      const platformList = Object.keys(platforms);
+      const opt = {};
+      for (let i = 0; i < platformList.length; i += 1) {
+        const platform = platformList[i];
+        const addKey = `binded_${platform}`;
+        opt[addKey] = true; // platform key exists only when binded
```
williamchong007: Use `.reduce` would be cleaner?

PR #902, opened 2018-12-11
Feature/google optimize

williamchong007:
@nwingt Please comment on both the implementation of `register-icon`, and the general way to do this kind of experiment


components/dialogs/AuthDialog.vue
```diff
@@ -32,7 +32,16 @@
...
-      <img :src="LikeCoinLogo">
+      <img
+        v-if="avatar"
+        :src="avatar"
+        class="auth-dialog__avatar auth-dialog__avatar--user"
+      >
+      <img
+        v-else
+        :src="LikeCoinLogo"
+        class="auth-dialog__avatar auth-dialog__avatar--likecoin"
```
nwingt: I prefer to name this class as `auth-dialog__logo`, with modifier `--avatar` and `--likecoin`.

I suggest to add a computed property i.e. `logoProps` which return
```javascript
{
  src: this.shouldShowFromUserIcon ? this.avatar : LikeCoinLogo,
  class: `auth-dialog__logo auth-dialog__logo--${this.shouldShowFromUserIcon ? 'avatar' : 'likecoin'}`,
}
```
so we can use `v-bind="logoProps"` with one ``

components/dialogs/AuthDialog.vue
```diff
@@ -291,6 +302,12 @@ export default {
         this.$t('AuthDialog.label.signInError')
       );
     },
+    shouldShowFromUserIcon() {
+      if (!this.$exp) return false;
+      const { name, $activeVariants } = this.$exp;
+      return name === 'register-icon'
+        && !!$activeVariants.find(variant => variant.name === 'from-user-avatar');
+    },
```
nwingt: I suggest to create a functional mixin for later use

e.g. 
```javascript
function abTestMixin(
  name,
  expName,
  varName,
) {
  return {
    computed: {
      [name]() {
        if (!this.$exp) return false;
        const { name, $activeVariants } = this.$exp;
        return name === expName
          && !!$activeVariants.find(variant => variant.name === varName);
      },
    },
  };
}

// AuthDialog.vue
...
mixins: [
  abTestMixin({
    shouldShowFromUserIcon,
    'register-icon',
    'from-user-avatar'
  })
],
...
```

components/dialogs/AuthDialog.vue
```diff
@@ -325,6 +342,18 @@ export default {
     },
   },
   async mounted() {
+    if (this.shouldShowFromUserIcon) {
```
nwingt: Is `shouldShowFromUserIcon` ready when executing `mounted()`?

williamchong007: `shouldShowFromUserIcon` is populated by nuxt plugin, thus should be always ready

components/dialogs/AuthDialog.vue
```diff
@@ -291,6 +301,12 @@ export default {
         this.$t('AuthDialog.label.signInError')
       );
     },
+    logoProps() {
+      return {
+        src: this.shouldShowFromUserIcon ? this.avatar : LikeCoinLogo,
```
nwingt: if avatar default is LikeCoinLogo, can simply return `src: this.avatar`?

PR #903, opened 2018-12-12
+ Add civic liker pre registration

mckingho:


constant/index.js
```diff
@@ -135,3 +135,5 @@ export const AVATAR_DEFAULT_PATH = 'https://static.like.co/likecoin_de-portrait.
 export const LIKE_BUTTON_POST_MESSAGE_TARGET_ORIGIN = IS_TESTNET ? '*' : 'https://button.like.co';
 
 export const WALLET_NOTICE_EXPIRY_INTERVAL = IS_TESTNET ? 300 : 604800; // In second
+
+export const PRE_REG_CIVIR_LIKER_END_DATE = '2019-01-03T08:00:00+0800';
```
nwingt: I prefer using timestamp in millisecond
I suggest to rename this constant as `CIVIC_LIKER_START_DATE`

PR #912, opened 2018-12-14
Integrate Civic Liker Pre-registration

nwingt:


server/api/users.js
```diff
@@ -1041,4 +1042,48 @@ router.put('/users/read/:id', jwtAuth('write'), async (req, res, next) => {
...
+    }
+
+    const { id } = req.params;
+    if (req.user.user !== id) {
+      res.status(401).send('LOGIN_NEEDED');
+      return;
+    }
+
+    const doc = await dbRef.doc(id).get();
+    if (!doc.exists) throw new Error('user not exists');
```
williamchong007: should use `USER_NOT_EXIST`

williamchong007:
@nwingt  I kept getting 
```
               <Root>
[Vue warn]: Invalid prop: custom validator check failed for prop "halo".

found in

---> <LcAvatar> at LcAvatar.vue
```
should this checker accept empty string, or should all default value be `none`

PR #925, opened 2018-12-18
+ Add csrf token in `application/x-www-urlencoded` endpoints

williamchong007:
This PR run the csrfToken() function in store server init, then compare against the token stored in user cookie

After we figure out how to seperate like.co api and api.like.co properly, we can enforce csrf token checking across all endpoint


store/modules/actions/api-wrapper.js
```diff
@@ -28,8 +28,8 @@ async function apiWrapper({ commit, dispatch }, promise, opt = {}) {
       || error.message
       || error;
     if (!slient) commit(types.UI_ERROR_MSG, errorMsg);
-    console.error(error);
-    throw new Error(errorMsg);
+    console.error(errorMode);
```
nwingt: Why changed to `errorMode`?

PR #926, opened 2018-12-18
! Add default value for `this.platforms` in `/_id/`

williamchong007:
due to `SocialMediaConnect` expect prop `platforms` to be required in all situations


williamchong007:
on second thought, any idea why here the 2 apis are not using `Promise.all()` ? @nwingt

williamchong007:
@nnkken @nwingt please have a look again

PR #927, opened 2018-12-19
Civic Liker Page and CTA

edmondyu:


nwingt:
Translation key has been updated

PR #929, opened 2018-12-19
* Adapt Civic Liker Trial quota

nwingt:


server/api/users.js
```diff
@@ -1054,23 +1056,37 @@ router.put('/users/:id/civic/trial', jwtAuth('write'), async (req, res, next) =>
...
+    const configDoc = await configRef.doc('civicLiker').get();
+    const preRegQuota = (configDoc.exists && configDoc.data().preRegQuota) || 0;
 
-    const updateObj = {
-      isPreRegCivicLiker: true,
-    };
+    const preRegQuery = await dbRef
+      .where('isPreRegCivicLiker', '==', true)
+      .limit(preRegQuota)
+      .get();
```
williamchong007: should update a `total count` value when there is a new civicliker, instead of query whole db for every read

server/api/users.js
```diff
@@ -1054,23 +1056,37 @@ router.put('/users/:id/civic/trial', jwtAuth('write'), async (req, res, next) =>
...
+    const preRegCount = preRegQuery.docs.length;
+
+    const updateObj = {};
+    if (preRegCount >= preRegQuota) {
+      updateObj.isOnPreRegCivicLikerWaitingList = true;
```
williamchong007: use a string field that act as a enum i.e.`PreRegCivicLikerStatus = 'waiting_list'` in case we want to do more judgement later?

server/api/users.js
nwingt: Now minus quota for each new Civic Liker

nwingt: Updated with total count

locales/client/ja.json
```diff
@@ -1236,17 +1237,15 @@
...
-    "trial": {
-      "register": "体験を開始する",
-      "registered": "無料体験中",
-      "freeNow": "無料体験中",
-      "freeToTry": "無料体験",
-      "willBeHeld": "2019年1月2日から31日まで"
-    }
+    ]
   },
   "CivicTrialPage": {
```
nnkken: Should this change to `CivicLikerTrial`?

PR #954, opened 2019-01-07
= Move async data fetching to `created()` hook, instead of `mounted()`

williamchong007:


williamchong007:
if we move it to `created()` hook it will also be run in server too, on hold

PR #961, opened 2019-01-10
Improve UX for legacy user

nwingt:


components/dialogs/AuthDialog.vue
```diff
@@ -452,9 +472,12 @@ export default {
           this.errorCode = 'FIREBASE_EMAIL_LINK_AUTH_NO_EMAIL';
           return;
 
-        case 'USER_REGISTER_ERROR':
         case 'EMAIL_ALREADY_USED':
         case 'USER_ALREADY_EXIST':
+          if (this.registerFailureCount < REGISTER_FAILURE_COUNT_MAX) {
+            this.registerFailureCount += 1;
```
williamchong007: (just discussion) why not do the popup logic here, but with a watcher? seems more complex

PR #981, opened 2019-01-24
+ Add trial event page

nwingt:
Depends on likecoin/likecoin-api-public#13


pages/in/civic/trial/_id.vue
```diff
@@ -0,0 +1,288 @@
...
+
+    let error = '';
+    if (isSubscribedCivicLiker) {
+      error = 'paid';
+    } else if (isCivicLikerTrial) {
+      error = 'joined';
+    } else {
+      try {
+        await apiGetCivicLikerTrialEventById(route.params.id);
+        return {};
```
williamchong007: add a comment to show this is a normal case exit

PR #991, opened 2019-01-30
+ Add display name under avatar in AuthDialog

nwingt:


locales/client/zh.json
```diff
@@ -658,6 +658,7 @@
     }
   },
   "AuthDialog": {
+    "supportCreator": "支持{name}",
```
williamchong007: why no space here between `{name}`?

nwingt: 中文唔洗space

nwingt: Updated

PR #998, opened 2019-02-08
 🚑 Do not set csrf token via cookie in `/in/embed` entry

williamchong007:
side effect: register and update user wont work if user SPA through embed to register window, which should never happen


williamchong007:
@nnkken please re-review, the original approach does not work at all

PR #1011, opened 2019-02-13
💄 Update text for different Civic Liker status

nwingt:


williamchong007:
all those `if (true) then return` cases, does that mean we should actually update `civicLikerStatus` to reflect them all in a var

nwingt:
> all those `if (true) then return` cases, does that mean we should actually update `civicLikerStatus` to reflect them all in a var

Maybe like this
- `waiting`
- `trial`
- `subscribed`
- `renewable`
- `expired`

williamchong007:
sounds good, but `subscribed` and `renewable` might overlap?

nwingt:
`subscribed` seems ambiguous 

How about this

* `waiting`
* `trial`
* `paid`
* `renewable`
* `expired`

PR #1027, opened 2019-02-21
✨ Set default P2P amount to USD0.25

nwingt:


pages/_id/index.vue
```diff
@@ -291,25 +292,40 @@ export default {
...
     }
-    return Promise.all([
+
+    const promises = [
       apiGetUserMinById(params.id),
       apiGetSocialListById(params.id).catch(() => ({})),
-    ]).then((res) => {
+    ];
+    if (!store.getters.getLikeCoinUsdNumericPrice) {
+      promises.push(store.dispatch('queryLikeCoinUsdPrice'));
```
williamchong007: who is actually firing this restful call here? server or client?

PR #1053, opened 2019-03-13
* Update registration error message

mckingho:


mckingho:
https://github.com/likecoin/likecoin-api-public/pull/22

PR #1059, opened 2019-03-18
🎨 Add email based LikerID prefill

williamchong007:


components/dialogs/AuthDialog.vue
```diff
@@ -872,6 +881,36 @@ export default {
...
+          try {
+            await apiGetUserMinById(tryName); // eslint-disable-line no-await-in-loop
+          } catch (err) {
+            if (err.response) {
+              if (err.response.status === 404) {
+                isIDAvailable = true;
+              }
+            }
+          }
+          tryName = `${suggestedName}${String(Math.ceil(Math.random() * 10000)).padStart(RANDOM_DIGIT_LENGTH, '0')}`;
```
nnkken: `tryName` is always changed even if `isIDAvailable` is true

nnkken: Also, should the harcoded `10000` depends on the value of `RANDOM_DIGIT_LENGTH`, say `Math.pow(10, RANDOM_DIGIT_LENGTH)`?

williamchong007: Updated to use`break` and `Math.floor` + `**`

PR #1063, opened 2019-03-20
🐛 Allow email unverified account to access tx history page

Aludirk:


williamchong007:
The commit message should describe what does the commit do, not description of an issue

PR #1080, opened 2019-04-10
🚸 Improve registration page UX by showing error details

nwingt:


components/dialogs/AuthDialogContent/Register.vue
```diff
@@ -84,9 +86,11 @@
               :title="$t('Register.form.error.emailFormat')"
               v-bind="$testID('RegisterForm-EmailField')"
               required
+              @focus="isFocusedEmailField = true"
+              @blur="isFocusedEmailField = false"
```
williamchong007: how about just call a `checkEmailError` method every `@change`?

components/dialogs/AuthDialogContent/Register.vue
```diff
@@ -185,13 +192,50 @@ export default {
     avatarSrc() {
       return this.avatarData || this.prefilledData.avatarURL;
     },
-    isFormValid() {
-      const isIdValid = new RegExp(LIKECOIN_ID_REGEX).test(this.likeCoinId);
-      const isEmailValid = (
+    isIdValid() {
+      return new RegExp(LIKECOIN_ID_REGEX).test(this.likeCoinId);
```
williamchong007: should reuse a same regex instance instead of new-ing `new RegExp(LIKECOIN_ID_REGEX)` everytime

PR #1099, opened 2019-04-18
💄 Add SignIn toggle in AuthDialog

williamchong007:


components/dialogs/AuthDialog.vue
```diff
@@ -90,7 +109,16 @@
           key="portal"
           class="auth-dialog__tab auth-dialog__tab--index"
         >
-          <signin-portal @submit="signInWithPlatform" />
+          <Transition
+            name="fade"
+            mode="out-in"
+          >
+            <signin-portal
+              :key="isSignIn.toString()"
```
nnkken: So the key is `"true" / "false"`?

components/dialogs/AuthDialog.vue
```diff
@@ -331,6 +353,7 @@ export default {
 
       errorCode: '',
       isSigningInWithEmail: false,
+      isSignIn: this.$route.query.login === '1' || false,
```
nnkken: In which case the ` || false` part is useful?

components/dialogs/AuthDialog.vue
williamchong007: Yes, since string seems more intuitive as key

PR #1121, opened 2019-05-07
⚡️ Reduce css size by PurgeCSS

williamchong007:


nuxt.config.js
```diff
@@ -281,4 +285,8 @@ if (process.env.INTERCOM_APPID) {
   );
 }
 
+if (process.env.NODE_ENV !== 'production') {
+  nuxtConfig.purgeCSS.enabled = true; // force purge in dev
```
nnkken: Seems the `enabled` option could be a function?
https://github.com/Developmint/nuxt-purgecss#enabled

williamchong007: I tried, and the param passed into it was always `undefined` when in dev mode

PR #1126, opened 2019-05-08
🎨 Move swiper into component

williamchong007:


util/SwiperDirectives.js
```diff
@@ -0,0 +1,118 @@
...
+        swiper.destroy && swiper.destroy()
+        delete vnode.context[instanceName]
+      }
+    }
+  }
+}
+/* End external source */
+/* eslint-enable */
+
+export default swiperDirective;
```
nwingt: should export `swiperDirective({})` for global options

PR #1129, opened 2019-05-08
🔒 Add sameSite in CSRF cookie

williamchong007:


williamchong007:
@nwingt @nnkken please review again

PR #1156, opened 2019-05-31
🚸 Try popup sign in if redirect failed

nwingt:


williamchong007:
Will user cancel trigger this popup?

nwingt:
closing the window should cancel

williamchong007:
i mean if user cancel in redirect version, a popup will trigger and they will have to press cancel again?

nwingt:
What do you mean by canceling redirect version? By clicking back in browser?

williamchong007:
(actually i just realised there is no cancel button in redirect...)
does it mean we cannot distinguish user cancel or (possible) bug case where user indeed just pressed back button?
not sure if it is considered bad ux if you get a popup everytime you press back

nwingt:
Yes that's the problem

PR #1157, opened 2019-05-31
Post-revamp of AuthDialog

nwingt:
Mainly about UI

### DIFF
--- a/components/dialogs/AuthDialog.vue
+++ b/components/dialogs/AuthDialog.vue
@@ -73,7 +73,7 @@
           <div
             v-if="currentTab === 'portal'"
             v-bind="tabProps"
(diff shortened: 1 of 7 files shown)


nwingt:
Better review with `Hide whitespace changes` option on

components/dialogs/AuthDialog.vue
```diff
@@ -588,20 +616,86 @@ export default {
...
+        case 'fade':
+        default:
+          this.$gsap.TweenLite.to(el, 1, {
+            opacity: 0,
+            onComplete,
+          });
+          break;
+      }
+    },
+    onTabBeforeEnter(el) {
```
nnkken: Where is this function used?

PR #1163, opened 2019-06-04
🐛 Fix querystring not parsed in redirect

nwingt:


williamchong007:
why `encodeURIComponent` was not needed before? Will adding it cause double encoding in end result? @nwingt

williamchong007:
btw why not use `$route.fullPath` or `$route.path`, and `window.location.href` might contains querystring then this might break, can detect `$route.query` then decided `&` or `?` if we dont want to parse + encode it again?

PR #1173, opened 2019-06-19
Introduce graphic AB test for SigninPortal

nwingt:


components/dialogs/AuthDialogContent/SignInPortal.vue
```diff
@@ -134,7 +136,7 @@ export default {
   },
   methods: {
     getExperimentLocale(key) {
-      if (this.shouldUseAltLocales && this.$te(`${key}-alternative`)) {
+      if (this.$te(`${key}-alternative`)) {
```
williamchong007: Should not default to alternative without checking, either default to key, or rename `shouldUseAltGraphic` to `shouldUseAltAsset` and check it first?

PR #1199, opened 2019-07-19
Start experiment on sign-in portal

nwingt:
Depends on likecoin/liker-land#152 and likecoin/likecoin-button#203


williamchong007:
probably should have a `/in/login` instead of `/in/register` defaulting to login, and needs a  `login=0` to actually register

PR #1202, opened 2019-07-24
🚸 Set AuthDialog default to login

nwingt:


components/dialogs/AuthDialog.vue
```diff
@@ -343,7 +343,7 @@ export default {
       errorCode: '',
       error: undefined,
       isSigningInWithEmail: false,
-      isSignIn: this.$route.query.login === '1',
+      isSignIn: this.$route.name === 'in-login',
```
williamchong007: isSignIn should be default, so condition should be `$route.name === 'in-register' || query.register === '1'`

PR #1212, opened 2019-08-02
🎨 Rewrite v-html part in tx history

williamchong007:


components/TransactionHistory.vue
```diff
@@ -119,10 +119,23 @@
...
                   <div
+                    v-for="value in [getValue(tx)]"
+                    :key="tx.id"
                     class="value"
-                    v-html="getValue(tx)"
-                  />
+                  >
+                    <!-- eslint-enable vue/valid-v-for -->
+                    <span class="lc-font-size-14">{{ value.integerPart }}</span>
+                    <span class="lc-font-size-12">{{ value.decimalPart }}</span>
```
nwingt: Is it missing a `.` between `integerPart` & `decimalPart`?

PR #1223, opened 2019-08-13
⚗️ Apply AB test result of sign in portal

nwingt:


nuxt.config.js
```diff
@@ -191,15 +191,15 @@ const nuxtConfig = {
       respectDoNotTrack: true,
     }],
     '@nuxtjs/sentry',
-    '@likecoin/nuxt-google-optimize',
+    // '@likecoin/nuxt-google-optimize',
   ],
   googleOptimize: {
```
williamchong007: Comment this whole block?

PR #1232, opened 2019-08-30
Add pending LIKE related pages

nwingt:


locales/client/en.json
```diff
@@ -824,7 +824,7 @@
   "In": {
     "tab": {
       "overview": "Overview",
-      "bonus": "Bonus",
+      "rewardRecords": "Reward Records",
```
nnkken: Why only `en` has this entry modified?

PR #1250, opened 2019-09-30
💄 Fine tune OAuth permission dialog

nwingt:
image


components/dialogs/OAuthPermissionDialog.vue
```diff
@@ -119,6 +128,9 @@ export default {
     providerLogo() {
       return (this.provider.avatar || DefaultProviderLogo);
     },
+    avatarHalo() {
```
nnkken: Unused in code?

nwingt: Fixed

PR #1255, opened 2019-10-08
Deploy matters branch with icon hidden

williamchong007:


components/settings/AuthConnectList.vue
```diff
@@ -169,7 +171,7 @@ export default {
 }
 
 .l-button {
-  @each $platform in google facebook twitter {
+  @each $platform in google facebook twitter matters {
     &--#{$platform} {
       background-color: map-get($platform-color-map, $platform);
```
nwingt: does `$platform-color-map have` `matters` in `~/assets/_variables.scss`?

williamchong007: no, not sure what should it be

nwingt: use `white` then

PR #1285, opened 2019-11-20
📱 Redirect mobile users to get app page

williamchong007:


locales/client/en.json
```diff
@@ -1633,6 +1633,13 @@
       }
     }
   },
+  "GetApp": {
+    "title": "Reinventing the Like, in your pocket",
+    "button": {
+      "downloadNow": "Download now",
+      "continueToWeb": "Visis without app"
```
nwingt: Visis

PR #1319, opened 2019-12-03
🐛 Fix migration dialog still appears when erc20 balance in below min …

williamchong007:
…amount
Fixes https://github.com/likecoin/erc-cosmos-migration/issues/32

### DIFF
--- a/store/modules/getters/user.js
+++ b/store/modules/getters/user.js
@@ -1,3 +1,6 @@
+import BigNumber from 'bignumber.js';
+import { ETH_MIGRATION_MIN_LIKECOIN_AMOUNT } from '~/constant';
+
 export const getUserInfo = state => state.user;
 
 export const getUserId = state => (state.user || {}).user;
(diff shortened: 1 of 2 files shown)


store/modules/getters/user.js
```diff
@@ -34,7 +37,8 @@ export const getUserLikeCoinAmountIsZero = state => (
 );
 
 export const getUserHasERC20LikeCoin = state => !!(state.likecoinAmountObject
-  && state.likecoinAmountObject.walletLIKE && state.likecoinAmountObject.walletLIKE !== '0');
+  && state.likecoinAmountObject.walletLIKE
+  && new BigNumber(state.likecoinAmountObject.walletLIKE)).gt(ETH_MIGRATION_MIN_LIKECOIN_AMOUNT);
```
nnkken: The condition in migration page is `toBN(value).lt(toBN(ETH_MIN_LIKECOIN_AMOUNT))`, so here should use `gte`

PR #1324, opened 2019-12-05
🐛 Fix authcore user stuck in register with email invalid char

williamchong007:


williamchong007:
@nwingt found a better place for the `isHideFilledInfo` shortcut, please review again

PR #1329, opened 2019-12-10
✨ Add error message for metamask register dialog error

williamchong007:


locales/client/en.json
```diff
@@ -1,4 +1,4 @@
-{
+ {
```
nwingt: Please delete this space

PR #1330, opened 2019-12-10
📈 Add Authcore register dialog analytics

williamchong007:


components/dialogs/AuthDialog.vue
```diff
@@ -451,14 +451,14 @@ export default {
...
+              delete query.register;
             } else {
-              delete query.login;
+              query.register = '1';
             }
           } else {
             // Remove show_login and login in query string
             delete query.show_login;
-            delete query.login;
+            delete query.register;
```
nnkken: not consistent with the comment

PR #1334, opened 2019-12-16
Minor UI updates

nwingt:


components/home/CosmosMigrationBanner.vue
```diff
@@ -0,0 +1,103 @@
...
+
+<script>
+import { mapGetters } from 'vuex';
+import { getMigrationSiteURL } from '@/util/api/api';
+
+export default {
+  name: 'cosmos-migration-banner',
+  computed: {
+    ...mapGetters([
+      'getUserInfo',
```
williamchong007: this component might be cleaner if we can just pass `user` string as prop, and call `getUserInfo ` from  parent instead

PR #1348, opened 2019-12-23
🚑 Mitigate email not fetched in some oauth register

williamchong007:


components/dialogs/AuthDialog.vue
```diff
@@ -855,10 +856,18 @@ export default {
       await this.setAuthCoreToken(accessToken);
       this.currentTab = 'loading';
       this.platform = 'authcore';
+      let user = currentUser;
+      if (!user || !user.profileName) {
+        try {
+          user = await this.fetchAuthCoreUser();
+        } catch (err) {
+          console.error(err);
```
nnkken: continue the execution with no user?

williamchong007: yes, the `user` is useful for frontend prefill purpose. As long as we have `id_token` and `access_token` we can handle authentication and email in backend.

PR #1352, opened 2020-01-08
Update home page & about page

nwingt:


components/home/OfficialChannelList.vue
```diff
@@ -0,0 +1,162 @@
...
+        },
+        {
+          id: 'matters',
+          link: 'https://matters.news/@likecoin/',
+        },
+        {
+          id: 'github',
+          link: 'https://github.com/likecoin/',
+          isHideBorder: true,
+        },
```
williamchong007: are the values in `LIKE_CO_PLATFORMS` of `constant` still used? we are currently using those for json ld object in index

PR #1361, opened 2020-01-23
🐛 Fix sourceURL was always overwritten by authcore.like.co

williamchong007:


components/dialogs/AuthDialog.vue
```diff
@@ -600,7 +600,7 @@ export default {
       if (window.sessionStorage) {
         // store data when first view page
         // restore data when it is lost. eg redirect sign in
-        if (this.sourceURL) {
+        if (this.sourceURL && !this.sourceURL.includes('like.co')) {
```
nnkken:
```JS
function shouldReadFromSession(sourceURL) {
  if (!sourceURL) {
    return false;
  }
  try {
    const url = new Url(sourceURL);
    return url.host.split('.').slice(-2).join('.') === 'like.co';
  } catch (err) {
    return false;
  }
}

...

if (shouldReadFromSession(this.sourceURL) {
  ...
} else {
  ...
}

```

PR #1370, opened 2020-02-05
Add newsletter feed

nwingt:


components/home/OfficialChannelList.vue
```diff
@@ -70,6 +101,47 @@ export default {
   components: {
     ImageGrid,
   },
+  data() {
+    return {
+      newsletters: [
+        {
+          title: 'LikeCoin Annual Report 2019',
+          pubDate: '2020-01-01 10:43:17',
```
nnkken: Better add timezone also?

nnkken: Actually I don't quite understand, since we just need the date string, why not just put the date string here, so we don't need to care about local time conversion and prevent inconsistency?

PR #1375, opened 2020-02-12
🍱 Add button sdk

williamchong007:


server/index.js
```diff
@@ -42,6 +42,11 @@ if (config.dev) {
   builder.build();
 }
 
+// explicit cache control on sdk resources
+app.use('/sdk', (req, res, next) => {
+  res.setHeader('Cache-Control', 'public, max-age=3600, s-maxage=3600, stale-if-error=604800, stale-while-revalidate=604800');
```
nnkken: For reference, the Facebook SDK entry point use this: `Cache-Control: public,max-age=1200,stale-while-revalidate=3600`.

williamchong007: we just copycat 1200 and 3600 then

PR #1385, opened 2020-03-04
💄 Clean up login related issues

williamchong007:


pages/in/getapp.vue
```diff
@@ -75,7 +75,7 @@ export default {
   }
 
   &__download-button {
-    margin: 20px;
+    margin: 20px 20px 10px 20px;
```
nwingt: Please change to padding

pages/in/getapp.vue
```diff
@@ -89,6 +89,8 @@ export default {
 
   &__about-likecoin-link,
   &__continue-button {
+    margin:10px;
```
nwingt: Please change to padding

pages/in/getapp.vue
williamchong007: should be margin, this is to decrease size for padding

PR #1387, opened 2020-03-10
🐛 Fix cannot edit external link

williamchong007:


ckxpress:
actually, i suggest to obsolete LikeCoin panel. it has little use (frankly i don't have stat, just guess), no future, and bring management overhead.

PR #1398, opened 2020-03-16
✨ Support comos MsgMultiSend and memo

williamchong007:
https://github.com/likecoin/like-co/pull/1394 was mis-merged, reopening
added memo and gas simulate support


pages/_id/index.vue
```diff
@@ -310,15 +319,20 @@ export default {
...
         fee,
         transfer_amount: transferAmount,
         transfer_to: transferTo,
         redirect_uri: redirectUri,
-        remark: remarks,
+        remarks,
       } = query);
     }
+    if (remarks) {
+      // TODO figure out if 255 is a good limit
```
nnkken: memo limit is 256, so it is actually OK

PR #1401, opened 2020-03-20
⚗ End position exp, start list style exp

williamchong007:


components/dialogs/AuthDialog.vue
```diff
@@ -92,7 +92,7 @@
               :is-sign-in="isSignIn"
               :language="getCurrentLocale"
               :redirect-url="getAuthCoreRedirectUrl"
-              :social-login-pane-style="socialLoginPanePosition"
+              :socialLoginPaneOption="socialLoginPaneOption"
```
nnkken: The props name is changed?

williamchong007: yes, 2 diff prop, but probably should be named with `-` instead of camel

PR #1406, opened 2020-03-26
🚑 Do not transpile authcore and secretd

williamchong007:
This fixs secretd-js cannot import bignumber.js properly due to transpile

Previously I encountered a `unexpected token "export"` error after upgrading authcore, so I added authcore to transpile
It seems that that error is gone now


williamchong007:
Seems not working if built prod... closing this for now

PR #1408, opened 2020-03-26
📈 Add crisp sdk in background to collect event

williamchong007:


util/EventLogger.js
```diff
@@ -14,6 +14,81 @@ function digestMessage(message) {
...
+    }
+    if (primaryPhone) $crisp.push(['set', 'user:phone', [primaryPhone]]);
+    const opt = [];
+    if (boot) opt.push(['LikeCoin', true]);
+    if (wallet) opt.push(['wallet', wallet]);
+    if (cosmosWallet) opt.push(['cosmos_wallet', cosmosWallet]);
+    if (LIKE) opt.push(['LIKE', LIKE]);
+    if (language) opt.push(['language', language]);
+    if (isAuthCore) opt.push(['binded_authcore', true]);
+    window.$crisp.push(['set', 'session:data', [opt]]);
```
nwingt: why dont just use `$crisp` here like above?

williamchong007:
> why Crisp API so ugly

I think it is a lazy trick for event queue before script is completely loaded

PR #1422, opened 2020-05-06
🚸 Show more verbose error for signin fail

williamchong007:


components/dialogs/AuthDialog.vue
```diff
@@ -410,7 +410,7 @@ export default {
           details: this.error.toString(),
         });
       }
-      return this.$t('AuthDialog.label.signInError');
+      return `${this.$t('AuthDialog.label.signInError')} - ${this.errorCode || ''}`;
```
nwingt, line 413: `登入時發生錯誤，請在右下對話框與我們聯絡 -` 
Maybe show some text instead of just ''? For visual only.

williamchong007: Not sure what would be a nice fallback, "unknown"?

nwingt: Up to u ~

williamchong007: How about
```
      let msg = this.$t('AuthDialog.label.signInError');
      if (this.errorCode) msg += ` - ${this.errorCode}`;
      return msg;
```

PR #1431, opened 2020-05-25
✨ Add matters redirect page

williamchong007:


pages/in/matters/redirect.vue
```diff
@@ -0,0 +1,7 @@
+<script>
+export default {
+  fetch({ redirect, query }) {
+    redirect(302, { name: 'in-getapp', query: { from: 'matters', ...query } });
```
nwingt, line 4: I think `from` is already reserved for LikeCoin button along with `referrer`? To prevent confusion, perhaps rename it to `source`?

PR #1443, opened 2020-06-18
📌 Pin whitepaper version

williamchong007:


pages/in/whitepaper.vue
```diff
@@ -87,18 +87,6 @@ const whitePaperSrc = [
...
-    src: '/in/likecoin-whitepaper-ko.pdf',
-  },
-  {
-    languageKey: 'ru',
-    src: '/in/likecoin-whitepaper-ru.pdf',
-  },
-  {
-    languageKey: 'vi',
-    src: '/in/likecoin-whitepaper-vi.pdf',
-  },
```
nwingt, line 101: Should remove cn & zh version as well?

williamchong007: we happen to have the right version of cn and zh white paper in git

nwingt: but it seems u remove in this commit ??

image
image

williamchong007: No, those bytes are the diff, this file is much larger than 20 bytes

nnkken: The diffs are very small (Bytes instead of MB), indicating that it is only updating the file instead of removing it.
Not sure why Github shows `100%`

PR #1446, opened 2020-07-03
✨ Add Liquid Quick Exchange to home page

nwingt:


components/LiquidQuickExchangeWidget/LiquidQuickExchangeWidget.vue
```diff
@@ -0,0 +1,114 @@
...
+            background: '#aaf1e7',
+          },
+        },
+      };
+      if (this.walletAddress) {
+        options.payout_settlement.input_parameters = {
+          account_key: {
+            type: 'WALLET_ADDRESS',
+            value: this.walletAddress,
+          },
```
williamchong007, line 85: Does it have differ input param for cosmos and eth wallet, does it matter in our case?

nwingt: Seems to only support Cosmos by now, the widget itself has its own address validation

PR #1465, opened 2020-09-03
💬 Update site title & desc

nwingt:


williamchong (commented):
@ckxpress should the zh json contains actual Chinese version of the description though

mileswong:
maybe we can keep it eng until we make chinese site.

williamchong (approved):

PR #1481, opened 2020-12-22
💬 Add default remarks for likepay

williamchong:
For distinguishing likepay tx


pages/in/widget/pay/index.vue
```diff
@@ -369,7 +369,7 @@ export default {
       state,
     } = query;
     let {
-      remarks,
+      remarks = 'via likepay',
```
nwingt, line 372: LIKE Pay?

nwingt (approved):

PR #1489, opened 2021-01-11
🌐 Handle language in qs for authenticated middleware

nwingt:


middleware/authenticated.js
```diff
@@ -43,15 +43,18 @@ export default function ({
...
+    register, from, referrer, language,
+  } = query;
   if (!store.getters.getUserIsRegistered) {
     let redirectPath = '/in/register';
-    const { register, from, referrer } = query;
     if (!process.server) {
       store.commit(USER_SET_AFTER_AUTH_ROUTE, route);
       setRouteToSessionStorage(route);
-      if (register === '1') redirectPath = `${redirectPath}?register=1`;
+      if (register === '1') redirectPath = `${redirectPath}?register=1&language=${language}`;
```
williamchong, line 54: Will language be set as `undefined` if it is not set? should we check for `language` exists or default it to `''` ?

williamchong (changes requested):

middleware/authenticated.js
nwingt: Added default

williamchong: I think it's better to not import the default locale into this file. Can we not handle language if it is not set in qs, and focus the handling of default locale in the i18n plugin file?

nwingt: Updated

williamchong (approved):
Hmm not sure why the old code specially split the case to only include all qs in `process.server`. But if it works fine including all qs in redirect should not be an issue

PR #1514, opened 2021-04-19
🐛 Fix wordpress tag parsing bug

shellyYG:


williamchong (approved):
Some issue still exists in the tags input from WordPress plugin/ISCN signing widget though

PR #1516, opened 2021-05-04
🐛 Display and check gas fee before sending in _id

williamchong:


nwingt (commented):
Should recalculate gas fee when the amount update?

williamchong:
> Should recalculate gas fee when the amount update?

I think gas fee is not related to the amount sent, but the number of targer or msg type, which is constant here

nwingt (approved):

PR #1526, opened 2021-07-06
♻️  Restructure request with Cosmos Native Packages to replace Lunie

shellyYG:


util/CosmosHelper.js
```diff
@@ -4,13 +4,16 @@ import {
   COSMOS_DENOM,
 } from '@/constant';
 import { timeout } from '@/util/misc';
+import { LcdClient } from '@cosmjs/launchpad';
 
 export const DEFAULT_GAS_PRICE = [{ amount: 1000, denom: 'nanolike' }];
 export const DEFAULT_ISCN_GAS_PRICE = [{ amount: 0, denom: 'nanolike' }];
 const COSMOS_RESTFUL_API = '/api/cosmos/lcd';
+const GET_API_ENDPOINT = 'https://node.taipei.like.co/';
```
williamchong, line 12: Should not hardcode this endpoint, just use the proxy in `COSMOS_RESTFUL_API`

util/cosmos/iscn.js
```diff
@@ -1,36 +1,29 @@
...
-let Cosmos;
-
 async function initISCNCosmos() {
   if (iscnApi) return;
-  ([Cosmos] = await Promise.all([
-    import(/* webpackChunkName: "web3" */ '@lunie/cosmos-api'),
-  ]));
-  if (Cosmos.default) Cosmos = Cosmos.default;
-  iscnApi = new Cosmos(ISCN_DEV_RESTFUL_API, ISCN_TESTNET_CHAIN_ID);
+  iscnApi = LcdClient.withExtensions({ ISCN_API_ENDPOINT });
```
williamchong, line 14: Similarly, should use `ISCN_DEV_RESTFUL_API`, if an absolute path is needed, should concat `EXTERNAL_URL` and `ISCN_DEV_RESTFUL_API`

williamchong (changes requested):

PR #1527, opened 2021-07-14
♻️ Replace lunie with Cosmjs Native Libraries

shellyYG:


util/cosmos/iscn.js
```diff
@@ -1,36 +1,29 @@
...
 import { ISCN_PUBLISHERS, ISCN_LICENSES } from './iscnConstant';
-import {
-  ISCN_TESTNET_CHAIN_ID,
-} from '@/constant';
 import { timeout } from '@/util/misc';
 import { apiPostISCNMessageForSign } from '../api/api';
 import { assertOk, queryTxInclusion } from './misc';
 
 const ISCN_DEV_RESTFUL_API = '/api/cosmos/iscn-dev/lcd';
+const EXTERNAL_URL = 'https://node.iscn-dev.like.co/';
```
williamchong, line 9: We can just use `ISCN_DEV_RESTFUL_API` instead of adding another `EXTERNAL_URL`

store/modules/actions/authCore.js
```diff
@@ -126,15 +127,41 @@ export async function fetchAuthCoreCosmosWallet({ state }) {
   return cosmosAddress;
 }
 
-export async function prepareAuthCoreCosmosTxSigner({ state }) {
-  if (!state.cosmosProvider) throw new Error('COSMOS_WALLET_NOT_INITED');
-  return async function signer(signMessage) {
-    const data = JSON.parse(signMessage);
-    const dataWithSign = await state.cosmosProvider.sign(data);
-    const signObject = dataWithSign.signatures[dataWithSign.signatures.length - 1];
```
williamchong, line 134: I think only  the part `return async function signer(signMessage) {` needs to be change, should not add `setupCosmosProvider` flow as it is already implement above as `initAuthCoreWalletService`, and the old flow does not handle reauth

store/modules/actions/authCore.js
```diff
@@ -1,7 +1,8 @@
 import { AuthCoreAuthClient } from 'authcore-js';
 import * as types from '@/store/mutation-types';
-import { AUTHCORE_API_HOST } from '@/constant';
+import { AUTHCORE_API_HOST, IS_TESTNET } from '@/constant';
 
+const authcoreURL = IS_TESTNET ? 'https://likecoin-integration-test.authcore.io' : 'https://authcore.like.co';
```
williamchong, line 5: Why do we need this now if we don't need it before?

store/modules/actions/payment.js
```diff
@@ -51,7 +50,6 @@ export async function sendCosmosPayment(
     commit(types.UI_SET_HIDE_TX_DIALOG_ACTION, !showDialogAction);
     commit(types.PAYMENT_SET_PENDING_HASH, txHash);
     commit(types.PAYMENT_SET_PENDING_TX_INFO, { from, to, value });
-    if (isWait) await included();
```
williamchong, line 54: What if we want to skip waiting for tx confirm (`isWait: false`)?  Does ` await transferCosmos` wait for tx confirm? Can we not wait for confirm according to `isWait` ?

util/CosmosHelper.js
```diff
@@ -122,66 +134,84 @@ export async function getTransferInfo(txHash, opt) {
...
+}, signer) {
   const amount = LIKEToAmount(value);
-  const msgPromise = api.MsgSend(from, { toAddress: to, amounts: [amount] });
-  return sendTx(msgPromise, signer, { memo, simulate });
+  const fee = {
+    amount: [{
+      denom: 'nanolike',
+      amount: '1000',
+    }],
+    gas: '44000',
```
williamchong, line 156: Should not hardcode here, should hardcode as constant like `DEFAULT_GAS_PRICE`

util/CosmosHelper.js
```diff
@@ -122,66 +134,84 @@ export async function getTransferInfo(txHash, opt) {
...
+  };
+  const defaultGasPrice = GasPrice.fromString('1000nanolike');
```
williamchong, line 158: Why do we define both `fee` and `defaultGasPrice`? Which one is really used?

util/CosmosHelper.js
```diff
@@ -122,66 +134,84 @@ export async function getTransferInfo(txHash, opt) {
...
+  }
+  const sendMsg = {
+    type: 'cosmos-sdk/MsgSend',
+    value: {
+      from_address: from,
+      to_address: to,
+      amount: [amount],
+    },
+  };
+  const broadcastedTx = await signingCosmosClient.signAndBroadcast([sendMsg], fee, memo);
```
williamchong, line 170: Does this break if user select another gas price  in keplr UI,other than the hardcoded `fee` here?

util/CosmosHelper.js
```diff
@@ -122,66 +134,84 @@ export async function getTransferInfo(txHash, opt) {
...
   });
-  const message = {
+  const fee = {
+    amount: [{
+      denom: 'nanolike',
+      amount: '1000',
+    }],
+    gas: '44000',
+  };
+  const defaultGasPrice = GasPrice.fromString('1000nanolike');
```
williamchong, line 200: Similarly, should use constant and not hardcode here, also no idea which one of `defaultGasPrice` and `fee` is actually used

util/api/api.js
```diff
@@ -2,7 +2,7 @@ import axios from '~/plugins/axios';
 import { IS_TESTNET, EXTERNAL_HOSTNAME } from '@/constant';
 
 const LIKECOIN_API_BASE = IS_TESTNET ? 'https://api.rinkeby.like.co' : 'https://api.like.co';
-const LIKECOIN_MISC_API_BASE = `https://${EXTERNAL_HOSTNAME}`;
+const LIKECOIN_MISC_API_BASE = IS_TESTNET ? 'https://rinkeby.like.co' : `https://${EXTERNAL_HOSTNAME}`;
```
williamchong, line 5: We should check `TEST_MODE` instead of `IS_TESTNET` since this issue only happen in local dev

williamchong (changes requested):

store/modules/actions/authCore.js
shellyYG: because I have "setupCosmosProvider" flow. If I delete "setupCosmosProvider", then this line could be deleted too! will work on deletion together

util/CosmosHelper.js
shellyYG: defaultGasPrice is used due to signingCosmosClient need to be in GasPrice form. 
However, to not repeat, I will extract fee value for defaultGasPrice.

store/modules/actions/payment.js
shellyYG: true. Will add included back.

util/CosmosHelper.js
shellyYG: No, tested both flow: authcore & Keplr. with send(==1) & sendMultiple(>1 receiver). Gas price will increase based on the chosen gas price by user.

williamchong (changes requested):

util/CosmosHelper.js
```diff
@@ -122,66 +131,79 @@ export async function getTransferInfo(txHash, opt) {
...
+  tos.forEach((target, index) => {
     outputs.push({
       address: target,
       coins: [amounts[index]],
     });
   });
-  const message = {
+  const fee = {
+    amount: DEFAULT_GAS_PRICE,
+    gas: '88000', // need to be higher than 44000 or error: out of gas in location: WriteFlat will happen
```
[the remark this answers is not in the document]
shellyYG: ok! will put them in '@/const'

util/CosmosHelper.js
```diff
@@ -122,66 +134,80 @@ export async function getTransferInfo(txHash, opt) {
...
+  tos.forEach((target, index) => {
     outputs.push({
       address: target,
       coins: [amounts[index]],
     });
   });
-  const message = {
+  const fee = {
+    amount: DEFAULT_GAS_PRICE,
+    gas: HEAVY_GAS,
```
williamchong, line 194: Can we estimate gas by using length of recipient? i.e. `tos.length * BASIC_GAS` ?

pages/_id/index.vue
```diff
@@ -395,17 +396,15 @@ export default {
...
-          value: this.amount,
-          memo: this.remarks,
-        },
-        null,
-        { simulate: true },
-      );
+      const fee = {
+        amount: [{
+          denom: COSMOS_DENOM,
+          amount: 1000,
```
williamchong, line 402: Is this `1000` related to `DEFAULT_GAS_PRICE` ? or `DEFAULT_GAS_PRICE * gas` ?

pages/in/widget/pay/index.vue
```diff
@@ -609,37 +609,32 @@ export default {
...
-          },
-          null,
-          { simulate: true },
-        ));
+        const fee = {
+          amount: [{
+            denom: COSMOS_DENOM,
+            amount: 1000,
+          }],
+          gas: parseInt(BASIC_GAS, 10),
```
williamchong, line 624: This is `isMultiSend` case, should the gas be `HEAVY_GAS` ? Also these gas estimate function should be refactored into cosmos helper

pages/in/widget/pay/index.vue
```diff
@@ -609,37 +609,32 @@ export default {
...
+        };
+        ({ gas } = fee);
```
williamchong, line 636: these gas estimate function should be refactored into cosmos helper

store/modules/actions/authCore.js
```diff
@@ -127,14 +127,19 @@ export async function fetchAuthCoreCosmosWallet({ state }) {
...
 export async function prepareAuthCoreCosmosTxSigner({ state }) {
+  let cosmosProvider;
   if (!state.cosmosProvider) throw new Error('COSMOS_WALLET_NOT_INITED');
-  return async function signer(signMessage) {
-    const data = JSON.parse(signMessage);
-    const dataWithSign = await state.cosmosProvider.sign(data);
-    const signObject = dataWithSign.signatures[dataWithSign.signatures.length - 1];
+  if (state.cosmosProvider) {
+    cosmosProvider = state.cosmosProvider; // eslint-disable-line prefer-destructuring
+  }
```
williamchong, line 134: since we will throw `COSMOS_WALLET_NOT_INITED` above, `state.cosmosProvider` always exists and not need for checking

constant/index.js
```diff
@@ -24,7 +24,10 @@ export const BIGDIPPER_HOST = IS_TESTNET ? 'https://taipei.likecoin.bigdipper.li
 
 export const COSMOS_WALLET_PREFIX = "m/44'/118'/0'/0";
 
-export const COSMOS_CHAIN_ID = IS_TESTNET ? 'likechain-testnet-taipei-1' : 'likecoin-chain-sheungwan';
+export const COSMOS_CHAIN_ID = IS_TESTNET ? 'likecoin-testnet-taipei-2' : 'likecoin-chain-sheungwan';
+
+export const BASIC_GAS = '44000';
```
williamchong, line 29: should be more explicit, like `BASIC_TRANSFER_GAS`

util/CosmosHelper.js
```diff
@@ -111,7 +122,8 @@ export async function getTransferInfo(txHash, opt) {
       _amount: amount,
     };
   }
-  const isFailed = (code && code !== '0') || !success;
+  const fullLogs = JSON.parse(rawLog.slice(1, rawLog.length - 1));
```
williamchong, line 125: Why use `.slice()`? If `rawLog` is an JSON array, we should parse full string and use `fullLogs[0]` below

util/CosmosHelper.js
```diff
@@ -122,66 +134,80 @@ export async function getTransferInfo(txHash, opt) {
...
+      to_address: to,
+      amount: [amount],
+    },
+  };
+
+  const broadcastedTx = await signingCosmosClient.signAndBroadcast([sendMsg], fee, memo);
+  return {
+    txHash: broadcastedTx.transactionHash,
+    gas: fee.gas,
+    gasPrices: fee.amount,
```
williamchong, line 172: Since `gasPrice` is changable in Keplr, should this `gasPrice` be a result from signed transaction, instead of hardcode fee?

util/CosmosHelper.js
```diff
@@ -190,29 +216,11 @@ async function MsgSendMultiple(
...
-      if (memo && memo.length) base += memo.length * 100;
-      return Math.floor(base * 1.2);
-    },
-    send: (
-      { gas, gasPrices = DEFAULT_GAS_PRICE, memo = undefined },
-      signer,
-    ) => api.send(senderAddress, { gas, gasPrices, memo }, message, signer),
+    txHash: broadcastedTx.transactionHash,
+    gas: fee.gas,
+    gasPrices: fee.amount,
```
williamchong, line 223: Since `gasPrice` is changable in Keplr, should this `gasPrice` be a result from signed transaction, instead of hardcode fee?

util/Keplr.js
```diff
@@ -134,14 +134,19 @@ class Keplr {
...
+          signed: signed.signed,
+          signature: {
+            signature: signed.signature.signature,
+            sequence: signed.signed.sequence,
+            account_number: signed.signed.account_number,
+            pub_key: signed.signature.pub_key,
+          },
+        };
+      },
     };
```
williamchong, line 150: Would this work? `this.signer` should already be an compatible signer without extra code
```return this.signer```

**Review by williamchong — CHANGES_REQUESTED** [review]

File: util/CosmosHelper.js (hunk 17 above)
Comment (shellyYG) [reply util/CosmosHelper.js:125]: oh! I misunderstood that JSON.parse can only be used in object.

File: store/modules/actions/authCore.js [hunk 21]
```diff
@@ -127,14 +127,17 @@ export async function fetchAuthCoreCosmosWallet({ state }) {
...
 export async function prepareAuthCoreCosmosTxSigner({ state }) {
+  let cosmosProvider;
   if (!state.cosmosProvider) throw new Error('COSMOS_WALLET_NOT_INITED');
-  return async function signer(signMessage) {
-    const data = JSON.parse(signMessage);
-    const dataWithSign = await state.cosmosProvider.sign(data);
-    const signObject = dataWithSign.signatures[dataWithSign.signatures.length - 1];
+  ({ cosmosProvider } = state); // eslint-disable-line prefer-const
+
+  if (cosmosProvider) {
```
williamchong, line 134: Why not just `const {cosmosProvider} = state` and skip the if part?

util/CosmosHelper.js
```diff
@@ -122,66 +133,92 @@ export async function getTransferInfo(txHash, opt) {
...
     type: 'cosmos-sdk/MsgMultiSend',
     value: {
       inputs: [
         {
-          address: senderAddress,
+          address: from,
           coins: [
             {
-              denom: COSMOS_DENOM,
+              denom: 'nanolike',
```
williamchong, line 221: Why not use `COSMOS_DENOM`?

util/CosmosHelper.js
```diff
@@ -1,25 +1,34 @@
...
-    import(/* webpackChunkName: "web3" */ '@likecoin/cosmos-api'),
-  ]));
-  if (Cosmos.default) Cosmos = Cosmos.default;
-  api = new Cosmos(COSMOS_RESTFUL_API, COSMOS_CHAIN_ID);
+async function initSigningCosmosClient(url, signerAddress,
+  signer, gasPrice, gasLimits, broadcastMode) {
+  if (signingCosmosClient) return;
+  signingCosmosClient = new SigningCosmosClient(url, signerAddress,
+    signer, gasPrice, gasLimits, broadcastMode);
 }
```
williamchong, line 32: which part of this function requires it to be `async` ?

util/CosmosHelper.js
```diff
@@ -1,25 +1,34 @@
...
+
+let cosmosClient;
+let signingCosmosClient;
 
-let Cosmos;
-let api;
+async function initCosmosClient() {
+  if (cosmosClient) return;
+  cosmosClient = new CosmosClient(COSMOS_RESTFUL_API);
+}
```
williamchong, line 25: which part of this function requires it to be `async` ?

util/Keplr.js
```diff
@@ -132,11 +132,10 @@ class Keplr {
...
     await this.checkIfInited();
-    const signerInstance = this.signer;
     const address = this.internalGetWalletAddress();
     return {
       signAmino: async (_, data) => {
-        const { signatures, ...signed } = await signerInstance.sign(address, data);
+        const { signatures, ...signed } = await this.signer.sign(address, data);
         return {
           signed: signed.signed,
           signature: {
```
williamchong, line 141: Can we skip the whole `signAmino` part and just return `this.signer`?
I think this.signer already has a built-in `signAmino()`

constant/index.js
```diff
@@ -24,7 +24,10 @@ export const BIGDIPPER_HOST = IS_TESTNET ? 'https://taipei.likecoin.bigdipper.li
 
 export const COSMOS_WALLET_PREFIX = "m/44'/118'/0'/0";
 
-export const COSMOS_CHAIN_ID = IS_TESTNET ? 'likechain-testnet-taipei-1' : 'likecoin-chain-sheungwan';
+export const COSMOS_CHAIN_ID = IS_TESTNET ? 'likecoin-testnet-taipei-2' : 'likecoin-chain-sheungwan';
+
+export const BASIC_TRANSFER_GAS = '44000';
+export const HEAVY_GAS = '88000';
```
williamchong, line 30: Do we still use `HEAVY_GAS` now?

williamchong (changes requested):

util/CosmosHelper.js
shellyYG: true. Changed back to COSMOS_DENOM.

util/CosmosHelper.js
shellyYG: true. changed to only function.

constant/index.js
shellyYG: no. will delete HEAVY_GAS.

store/modules/actions/authCore.js
```diff
@@ -127,10 +127,8 @@ export async function fetchAuthCoreCosmosWallet({ state }) {
 }
 
 export async function prepareAuthCoreCosmosTxSigner({ state }) {
-  let cosmosProvider;
   if (!state.cosmosProvider) throw new Error('COSMOS_WALLET_NOT_INITED');
-  ({ cosmosProvider } = state); // eslint-disable-line prefer-const
-
+  const { cosmosProvider } = state;
   if (cosmosProvider) {
```
williamchong, line 132: Please remove the if part for `if (cosmosProvider) {` since `if (!state.cosmosProvider)` already guard this case, just use `cosmosProvider` directly

williamchong (changes requested):

williamchong (approved):

PR #1528, opened 2021-07-20
💬 Change wallet exists error string

williamchong:


edmondyu (changes requested):
Should add a link to CS https://go.crisp.chat/chat/embed/?website_id=5c009125-5863-4059-ba65-43f177ca33f7

nwingt (approved):

PR #1532, opened 2021-07-30
🚧  Add blocking for chain upgrade

nwingt:


williamchong:
Also need to block likepay and iscn widget too

williamchong (approved):

PR #1533, opened 2021-08-03
✨ Enable stargate proto direct sign

shellyYG:


util/cosmos/network.js
```diff
@@ -39,13 +39,12 @@ const mainConfig = {
...
 const testnetConfig = {
-  id: 'likecoin-testnet-taipei-2', // Change test net to taipei-2
-  name: 'LikeCoin Taipei test chain',
-  apiURL: 'https://node.taipei2.like.co',
-  rpcURL: 'ws://node.taipei2.like.co:26657',
-  stakingWalletURL: 'http://likecoin-chain-taipei2.netlify.app/',
+  id: 'iscn-dev-chain-2',
+  name: 'LikeCoin ISCN Dev Chain', // TO-DO
+  apiURL: 'https://node.iscn-dev-2.like.co',
+  rpcURL: 'https://node.iscn-dev-2.like.co/rpc',
```
williamchong, line 46: Should be `https://node.iscn-dev-2.like.co/rpc/`

store/modules/actions/authCore.js
```diff
@@ -130,9 +132,24 @@ export async function prepareAuthCoreCosmosTxSigner({ state }) {
   if (!state.cosmosProvider) throw new Error('COSMOS_WALLET_NOT_INITED');
   const { cosmosProvider } = state;
   return {
-    signAmino: async (_, data) => {
-      const { signatures, ...signed } = await cosmosProvider.sign(data);
-      return { signed, signature: signatures[0] };
```
williamchong, line 135: Can we actually keep `signAmino` here without using it, and let cosmjs stargate use `signDirect` ?

util/CosmosHelper.js
```diff
@@ -7,28 +7,31 @@ import {
...
 
 export const DEFAULT_GAS_PRICE = [{ amount: '1000', denom: COSMOS_DENOM }];
 export const DEFAULT_GAS_PRICE_NUMBER = parseInt(DEFAULT_GAS_PRICE[0].amount, 10);
 export const DEFAULT_ISCN_GAS_PRICE = [{ amount: 0, denom: COSMOS_DENOM }];
-const COSMOS_RESTFUL_API = `${EXTERNAL_URL}/api/cosmos/lcd`;
-
-let cosmosClient;
-let signingCosmosClient;
+const COSMOS_RPC_API = `${EXTERNAL_URL}/api/cosmos/rpc`;
+const COSMOS_RESTFUL_API = 'https://node.iscn-dev-2.like.co/';
```
williamchong, line 23: Should just keep `COSMOS_RESTFUL_API` unchanged

util/CosmosHelper.js
```diff
@@ -48,94 +51,92 @@ export function amountToLIKE(likecoin) {
...
-      const {
-        value: {
-          amount: [a],
-          from_address: f,
-          to_address: t,
-        },
-      } = m;
-      amount.push(a);
-      from.push(f);
-      to.push(t);
```
williamchong, line 105: The old code actually handles multiple `MsgSend` via a for loop, can we implement it with similar logic?

util/CosmosHelper.js
```diff
@@ -48,94 +51,92 @@ export function amountToLIKE(likecoin) {
...
-    rawLog,
   } = txData;
-  const fullLogs = JSON.parse(rawLog);
+  const block = await stargateClient.getBlock(height);
+  const { header: { time: timestamp } } = block;
+
   return {
     ts: (new Date(timestamp)).getTime(),
-    isFailed: (code && code !== '0') || !fullLogs[0].success,
+    isFailed: (code && code !== '0'),
```
williamchong, line 68: Is `code` a string or a number in the new cosmjs getTx() result?

util/CosmosHelper.js
```diff
@@ -48,94 +51,92 @@ export function amountToLIKE(likecoin) {
...
     return {
       _from: from,
       _to: to,
-      _amount: amount,
+      _amount: amounts,
     };
   }
-  const fullLogs = JSON.parse(rawLog.slice(1, rawLog.length - 1));
-  const isFailed = (code && code !== '0') || !fullLogs.success;
+  const isFailed = (code && code !== '0');
```
williamchong, line 127: Is `code` a string or a number in the new cosmjs getTx() result?

williamchong (changes requested):

util/CosmosHelper.js
shellyYG: it's now a number. Will change it to !== 0

util/CosmosHelper.js
shellyYG: it's now a number. Will change it to !== 0

util/CosmosHelper.js
shellyYG: ok added multiple messages support.

williamchong (changes requested):

util/CosmosHelper.js
```diff
@@ -118,9 +118,21 @@ export async function getTransferInfo(txHash, opt) {
...
+          if (Array.isArray(t)) {
+            to = to.concat(t);
+          } else {
+            to.push(t);
+          }
+          if (Array.isArray(a)) {
+            amounts = amounts.concat(a);
+          } else {
+            amounts.push(a);
+          }
```
williamchong: Can we merge the `inputs.length > 1` part and `Array.isArray()` part into one block so that it looks cleaner?
```
if (inputs.length > 1) {
  from = from.concat(inputs.map(i => i.address));
} else {
  from.push(inputs[0].address)
}
```

williamchong (changes requested):

PR #1536, opened 2021-08-11
🐛 Fix gas calculation bug by multiplying gas price

shellyYG:


pages/in/widget/pay/index.vue
```diff
@@ -620,7 +621,8 @@ export default {
       } else {
         ({ gas } = await calculateCosmosGas(this.toUsers));
       }
-      this.gasFee = new BigNumber(gas).dividedBy(1e9).toFixed();
+      this.gasFee = new BigNumber(gas)
+        .multipliedBy(parseInt(DEFAULT_GAS_PRICE[0].amount, 10)).dividedBy(1e9).toFixed();
```
williamchong, line 625: Just use the new feeAmount would do?

williamchong (changes requested):

williamchong (approved):

PR #1540, opened 2021-08-19
🐛 Fix creator avatar and link parsing in iscn view page

williamchong:


pages/in/tx/iscn/_id/index.vue
```diff
@@ -396,6 +396,25 @@ export default {
     totalStakeholdersShares() {
       return this.stakeholders.reduce((t, s) => t + s.rewardProportion, 0);
     },
+    creatorLikerId() {
+      if (this.creatorId.startsWith(EXTERNAL_URL)) {
+        this.creatorId.replace(EXTERNAL_URL, '');
```
nnkken, line 401: `String.prototype.replace` is not in-place, so this line has no effect

pages/in/tx/iscn/_id/index.vue
```diff
@@ -396,6 +396,25 @@ export default {
...
+    isCreatorIdURL() {
+      try {
+        const res = new URL(this.creatorId);
+        return !!res;
+      } catch (err) {
+        return false;
+      }
+    },
+    creatorLink() {
+      if (this.creatorLikerId) return this.getCreatorPortfolioURL(this.creatorLikerId);
```
nnkken, line 414: `this.creatorLikerId` always return `''` so this will never run?

pages/in/tx/iscn/_id/index.vue
```diff
@@ -471,13 +490,15 @@ export default {
...
-          apiGetUserMinById(this.creatorId).catch(() => ({})),
-        ]);
-        if (creatorData && creatorData.data) {
-          this.creatorName = creatorData.data.displayName || creatorData.data.user;
-          this.creatorAvatar = creatorData.data.avatar;
-          this.creatorAvatarHalo = UserUtil.getAvatarHaloType(creatorData.data);
+        if (this.creatorLikerId) {
+          const [creatorData] = await Promise.all([
+            apiGetUserMinById(this.creatorLikerId).catch(() => ({})),
+          ]);
```
nnkken, line 496: Why `Promise.all` at the first place if there is only one entry...

pages/in/tx/iscn/_id/index.vue
nnkken: Just got the idea, maybe you want `return this.creatorId.replace(EXTERNAL_URL, '');`?
The name `EXTERNAL_URL` is quite misleading in this context, since the logic is `if it starts with our "external URL", that means it's something internal and is a Liker ID`

williamchong: `EXTERNAL_URL` is computed from `EXTERNAL_HOSTNAME` of the site though..

pages/in/tx/iscn/_id/index.vue
williamchong: Should return a liker id after the fix above

shellyYG (approved):

nnkken (approved):

PR #1544, opened 2021-08-20
✨ Add ISCN gas limit calculation for big size tx

shellyYG:


shellyYG:
used another PR: https://github.com/likecoin/like-co/pull/1548

PR #1545, opened 2021-08-21
🐛 Fix Keplr wallet source not correctly inited problem

shellyYG:


shellyYG:
Use another way to solve this bug: https://github.com/likecoin/like-co/pull/1547

PR #1546, opened 2021-08-23
🚚 Redirect home page to about.like.co

nwingt:


pages/index.vue
```diff
@@ -1,738 +1,9 @@
...
-        // eslint-disable-next-line no-console
-        console.error('Quick Exchange Widget Logging Error:', error);
-      }
-    },
-    onQExError(errors) {
-      // eslint-disable-next-line no-console
-      console.error('Quick Exchange Widget Errors:', errors);
-    },
+  fetch({ redirect }) {
+    redirect(301, getLikeCoAboutURL());
```
williamchong, line 6: Would 302 be safer in case we want to change again later?

williamchong (changes requested):

williamchong (approved):

PR #1549, opened 2021-08-23
🐛 Fix loginByCosmosWallet keplr not dispatch walletSource issue

shellyYG:


williamchong:
Fixed by https://github.com/likecoin/like-co/pull/1547

PR #1553, opened 2021-09-01
🐛 Fix calculateISCNTotalFee gas not align with signISCNTx gas

shellyYG:


nwingt:
Need backward compatibility?

williamchong (approved):

williamchong:
> Need backward compatibility?

I think it is purely a bug, `title` was changed to `name` in actual signing and payload

PR #1554, opened 2021-09-01
♻️ Add gas price for ISCN tx to be explicit

shellyYG:


util/cosmos/iscn/sign.js
```diff
@@ -215,7 +214,7 @@ export async function estimateISCNTxGas(tx) {
   const gasUsedEstimation = byteSize.multipliedBy(GAS_ESTIMATOR_SLOP).plus(interceptWithBuffer);
   return {
     gasFee: {
-      amount: [{ amount: gasUsedEstimation.toFixed(0, 0), denom: 'nanolike' }],
+      amount: [{ amount: gasUsedEstimation.multipliedBy(DEFAULT_GAS_PRICE_NUMBER).toFixed(0, 0), denom: 'nanolike' }],
       gas: gasUsedEstimation.toFixed(0, 0),
```
williamchong, line 218: is nanolike a const defined as COSMOS_DENOM?

williamchong (changes requested):

util/cosmos/iscn/sign.js
shellyYG: true! will change this.

williamchong (approved):

PR #1563, opened 2021-11-18
♻️ Allow like pay widget pop up and communicate back to wordpress

shellyYG:


pages/in/widget/pay/index.vue
```diff
@@ -714,8 +727,17 @@ export default {
         if (state) url.query.state = state;
         if (remarks) url.query.remarks = remarks;
         if (success !== undefined) url.query.success = success;
-        url.set('query', url.query);
-        window.location.href = url.toString();
+        if (this.opener) {
+          const message = JSON.stringify({
+            action: 'TX_FEE_SUBMITTED',
```
williamchong, line 732: Since this is a generic like pay widget, just `TX_SUBMITTED` would done

pages/in/widget/pay/index.vue
```diff
@@ -701,7 +714,7 @@ export default {
       }
     },
     async postTransaction({ txHash, error } = {}) {
-      if (this.redirectUri) {
+      if (this.opener || this.redirectUri) {
```
williamchong, line 717: What happen in this part if `this.redirectUri` is not defined?

pages/in/widget/pay/index.vue
```diff
@@ -714,8 +727,17 @@ export default {
...
+        if (this.opener) {
+          const message = JSON.stringify({
+            action: 'TX_FEE_SUBMITTED',
+            data: url.query,
+          });
+          this.windowOpener.postMessage(message, this.redirectOrigin);
+          window.close();
+        } else {
+          url.set('query', url.query);
+          window.location.href = url.toString();
```
williamchong, line 739: I think it is better to make sure `redirectUri` is well defined and is inside users' white list before redirecting

williamchong (changes requested):

pages/in/widget/pay/index.vue
shellyYG: True, then `const url = new URL(this.redirectUri, true);` would have error. Need to change this.

pages/in/widget/pay/index.vue
shellyYG: Should we just separate it like iscn widget?

williamchong: or you can mark a flag in the asyncData block where we check the whitelist already (which this PR skipped)

pages/in/widget/pay/index.vue
```diff
@@ -439,40 +440,50 @@ export default {
...
+      let isRedirectURLWhiteListed = false;
       if (agentUser && !agentUser.cosmosWallet) {
-        error({ statusCode: 400, message: 'VIA_USER_HAS_NO_WALLET' });
+        redirectWhitelistError = redirectWhitelistError || { statusCode: 400, message: 'VIA_USER_HAS_NO_WALLET' };
       }
-
       if (toUsers.some(u => !u.cosmosWallet)) {
-        error({ statusCode: 400, message: 'RECEIPIENT_HAS_NO_WALLET' });
+        redirectWhitelistError = redirectWhitelistError || { statusCode: 400, message: 'RECEIPIENT_HAS_NO_WALLET' };
       }
```
williamchong, line 450: Please do not modify these error unrelated to `redirectUri`, just throw them directly

pages/in/widget/pay/index.vue
```diff
@@ -439,40 +440,50 @@ export default {
           paymentRedirectWhiteList,
         };
       });
+      let redirectWhitelistError = null;
+      let isRedirectURLWhiteListed = false;
```
williamchong, line 444: Better move these to below, above `if (redirectUri)` block, since that is the part where we handle all redirect related logic

pages/in/widget/pay/index.vue
```diff
@@ -439,40 +440,50 @@ export default {
...
+            redirectWhitelistError = redirectWhitelistError || { statusCode: 400, message: 'USER_NOT_SETUP_PAYMENT_REDIRECT' };
           }
           if (userWhiteList.length && !userWhiteList.includes(redirectUri)) {
-            error({ statusCode: 400, message: 'REDIRECT_URI_NOT_WHITELIST' });
+            redirectWhitelistError = redirectWhitelistError || { statusCode: 400, message: 'REDIRECT_URI_NOT_WHITELIST' };
+          }
+          if (userWhiteList.length && userWhiteList.includes(redirectUri)) {
+            isRedirectURLWhiteListed = true;
           }
         }
```
williamchong, line 481: I think we can just write `isRedirectURLWhiteListed = !err` here, at the end of the `if (redirectUri)` block
and no need to explicitly check for `if (agentWhiteList.length && agentWhiteList.includes(redirectUri))` and `if (userWhiteList.length && userWhiteList.includes(redirectUri)`

pages/in/widget/pay/index.vue
```diff
@@ -439,40 +440,50 @@ export default {
...
-            error({ statusCode: 400, message: 'AGENT_NOT_SETUP_PAYMENT_REDIRECT' });
+            redirectWhitelistError = redirectWhitelistError || { statusCode: 400, message: 'AGENT_NOT_SETUP_PAYMENT_REDIRECT' };
           }
           if (agentWhiteList.length && !agentWhiteList.includes(redirectUri)) {
-            error({ statusCode: 400, message: 'REDIRECT_URI_NOT_WHITELIST' });
+            redirectWhitelistError = redirectWhitelistError || { statusCode: 400, message: 'REDIRECT_URI_NOT_WHITELIST' };
+          }
+          if (agentWhiteList.length && agentWhiteList.includes(redirectUri)) {
+            isRedirectURLWhiteListed = true;
           }
```
williamchong, line 464: No need to add a extra block just to set this flag, can handle this below

pages/in/widget/pay/index.vue
```diff
@@ -439,40 +440,50 @@ export default {
...
           if (userWhiteList.length && !userWhiteList.includes(redirectUri)) {
-            error({ statusCode: 400, message: 'REDIRECT_URI_NOT_WHITELIST' });
+            redirectWhitelistError = redirectWhitelistError || { statusCode: 400, message: 'REDIRECT_URI_NOT_WHITELIST' };
+          }
+          if (userWhiteList.length && userWhiteList.includes(redirectUri)) {
+            isRedirectURLWhiteListed = true;
           }
         }
       }
+      const shouldThrowRedirectWhitelistError = redirectUri && opener !== '1' && opener !== 1;
```
williamchong, line 483: I think `opener !== '1' && opener !== 1` here and the `opener && opener !== '0'` below can be refactor into one variable?

williamchong (changes requested):

williamchong (changes requested):

williamchong (approved):

PR #1565, opened 2021-11-29
✨ Support fingerprint as array for multiple ipfsHash values and arweaveId

shellyYG:


pages/in/widget/iscn/index.vue
```diff
@@ -128,11 +128,18 @@
...
+                  <ul id="ipfs=list">
+                    <li
+                      v-for="(item, index) in fingerprint"
+                      :key="index"
+                    >
+                      <a
+                        :href="ipfsURLs[index]"
+                        target="_blank"
+                        rel="noopener"
+                      > {{ item }} <br></a>
```
nwingt, line 140:
```suggestion
                      >{{ item }}</a>
```

nwingt (changes requested):

pages/in/widget/iscn/index.vue
```diff
@@ -128,11 +128,18 @@
                   {{ $t('ISCNWidget.label.fingerprint') }}
                 </div>
                 <div class="likepay-panel__section-meta-grid-item-value likepay-panel__section-meta-grid-item-value--fingerprint">
-                  <a
-                    :href="ipfsURL"
-                    target="_blank"
-                    rel="noopener"
-                  >{{ fingerprint }}</a>
+                  <ul id="ipfs=list">
```
williamchong, line 131: Please remove `ipfs=list` if not used

pages/in/widget/iscn/index.vue
```diff
@@ -233,6 +240,7 @@ export default {
     if (!Object.keys(query).length) {
       return redirect('https://docs.like.co/developer/international-standard-content-number-iscn/web-widget');
     }
+    fingerprint = fingerprint.split(',');
```
williamchong, line 243: I think you can use a new variable `fingerprints`, because after `split(',')` event a string without `,` will turns into a single element array

pages/in/widget/iscn/index.vue
```diff
@@ -297,8 +305,27 @@ export default {
       if (!window) return null;
       return window.opener;
     },
-    ipfsURL() {
-      return `https://ipfs.io/ipfs/${this.fingerprint}`;
+    fingerprintURLs() {
+      const fingerprintList = [];
+      for (let i = 0; i < this.fingerprint.length; i += 1) {
+        const fingerprintParts = this.fingerprint[i].split('://');
```
williamchong, line 311: Prefer `.map()`

util/cosmos/iscn/sign.js
```diff
@@ -107,7 +107,15 @@ function preformatISCNPayload(payload) {
   }
 
   const contentFingerprints = [];
-  if (fingerprint) contentFingerprints.push(`ipfs://${fingerprint}`);
+  if (fingerprint) {
+    if (Array.isArray(fingerprint)) {
+      for (let i = 0; i < fingerprint.length; i += 1) {
+        contentFingerprints.push(fingerprint[i]);
```
williamchong, line 113: How about `.concat()`?

pages/in/widget/iscn/index.vue
```diff
@@ -297,8 +305,27 @@ export default {
...
+        switch (fingerprintParts[0]) {
+          case 'ipfs':
+            url = `https://ipfs.io/ipfs/${fingerprintParts[1]}`;
+            break;
+          case 'ar':
+            url = `https://arweave.net/${fingerprintParts[1]}`;
+            break;
+          case 'hash':
+            url = `https://mainnet-node.like.co/txs/${fingerprintParts[1]}`;
+            break;
```
williamchong, line 322: I dont think we have a `hash` type that is a transaction hash, we do use file sha256 hash but it doesn't have a url
Please remove this case

pages/in/widget/iscn/index.vue
```diff
@@ -297,8 +305,27 @@ export default {
...
-      return `https://ipfs.io/ipfs/${this.fingerprint}`;
+    fingerprintURLs() {
+      const fingerprintList = [];
+      for (let i = 0; i < this.fingerprint.length; i += 1) {
+        const fingerprintParts = this.fingerprint[i].split('://');
+        let url = '';
+        switch (fingerprintParts[0]) {
+          case 'ipfs':
+            url = `https://ipfs.io/ipfs/${fingerprintParts[1]}`;
+            break;
```
williamchong, line 316: We need to backward support old plugin which directly send ipfs hash into `fingerprint` without `ipfs://`
Can we check if a fingprint starts with `Qm` and has a certain length, if yes assume it is ipfs?

util/cosmos/iscn/sign.js
williamchong: Also what if a old plugin version directly send ipfs hash without `ipfs://` ? This will break iscn formatting

williamchong (changes requested):

pages/in/widget/iscn/index.vue
```diff
@@ -197,7 +204,7 @@ export default {
   data() {
     return {
       isLoading: false,
-      fingerprint: '',
+      fingerprint: [],
```
williamchong, line 207: Better rename to `fingerprints`

pages/in/widget/iscn/index.vue
williamchong: Should also guard case where `fingerprint` is `undefined`

williamchong (changes requested):

pages/in/widget/iscn/index.vue
shellyYG: ok, proposed change:
1. use new variable fingerprint`s`
2. put the fingerprints definition after fingerprint to guard undefined case:
`if (!fingerprint) {
      return error({ statusCode: 400, message: 'INVALID_FINGERPRINT' });
    }
    const fingerprints = fingerprint.split(',');`

pages/in/widget/iscn/index.vue
shellyYG: Yes, as discussed, also add ipfsHash length === 46 as guard

pages/in/widget/iscn/index.vue
```diff
@@ -297,8 +305,26 @@ export default {
...
       return window.opener;
     },
-    ipfsURL() {
-      return `https://ipfs.io/ipfs/${this.fingerprint}`;
+    fingerprintURLs() {
+      const fingerprintList = this.fingerprints.map((f) => {
+        const fingerprintParts = f.split('://');
+        let url;
+        if (fingerprintParts[0].startsWith('Qm')) {
+          return f; // support old wordpress plugin
```
williamchong, line 313: Should we return ipfs.io link here?

util/cosmos/iscn/sign.js
```diff
@@ -106,8 +106,16 @@ function preformatISCNPayload(payload) {
...
-  const contentFingerprints = [];
-  if (fingerprint) contentFingerprints.push(`ipfs://${fingerprint}`);
+  let contentFingerprints = [];
+  if (fingerprints) {
+    if (Array.isArray(fingerprints)) {
+      if (fingerprints[0].startsWith('Qm') && fingerprints[0].length === 46) {
+        contentFingerprints.push(`ipfs://${fingerprints[0]}`);
+      } else {
+        contentFingerprints = contentFingerprints.concat(fingerprints);
+      }
```
williamchong, line 116: Should check each entry with for-loop instead of hardcoding `[0]`

williamchong (changes requested):

util/cosmos/iscn/sign.js
williamchong: In fact we should handle this logic in page level, page should convert old format into new format

pages/in/widget/iscn/index.vue
shellyYG: true 🌚

pages/in/widget/iscn/index.vue
```diff
@@ -236,6 +243,16 @@ export default {
...
       return error({ statusCode: 400, message: 'INVALID_FINGERPRINT' });
     }
+    let fingerprints = fingerprint.split(',');
+    fingerprints = fingerprints.map((f) => {
+      let contentFingerprint;
+      if (f.startsWith('Qm') && f.length === 46) {
+        contentFingerprint = `ipfs://${f}`; // support old wordpress plugin
+      } else {
+        contentFingerprint = f;
+      }
```
williamchong, line 253:
```suggestion
      let contentFingerprint = f;
      if (f.startsWith('Qm') && f.length === 46) {
        contentFingerprint = `ipfs://${f}`; // support old wordpress plugin
      }
```

williamchong (changes requested):

williamchong (approved):

PR #1566, opened 2021-12-08
✨ Add iscnId as returned value for ISCN widget

shellyYG:


pages/in/widget/iscn/index.vue
```diff
@@ -279,6 +280,7 @@ export default {
       opener: opener && opener !== '0',
       state,
       url,
+      blocking,
```
williamchong, line 283: should check with same logic as `opener`

williamchong (changes requested):

pages/in/widget/iscn/index.vue
shellyYG: ok! added on next commit

williamchong (approved):

PR #1597, opened 2022-03-10
Feature/iscn ar widget

williamchong:
- use postmessage instead of query string for more responsive UI
- integrate arweave api into the widget to abstract away the need for calling like.co api from caller

### DIFF
--- a/locales/client/cn.json
+++ b/locales/client/cn.json
@@ -945,13 +945,14 @@
         }
     },
     "ISCNARWidget": {
-    "transaction": {
-      "firstStep": "签署  (1/2)",
-      "secondStep": "签署  (2/2)",
-      "reminder": "请以你的钱包签署",
-      "retry": "重试",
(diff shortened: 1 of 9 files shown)


locales/client/cn.json
```diff
@@ -945,13 +945,14 @@
...
-    "transaction": {
-      "firstStep": "签署  (1/2)",
-      "secondStep": "签署  (2/2)",
-      "reminder": "请以你的钱包签署",
-      "retry": "重试",
-      "doNotCloseReminder": "请勿关闭此视窗",
-      "signAgain": "你还需要再签署一次才完成程序"
+        "transaction": {
+            "firstStep": "签署  (1/2)",
+            "secondStep": "签署  (2/2)",
```
nwingt, line 950: Why there is extra space before the bracket?

williamchong: No idea, can remove

PR #1615, opened 2022-04-08
🎨 Replace legacy LCD endpoints

WeiJun0827:


util/cosmos/misc.js
```diff
@@ -17,7 +17,7 @@ export function assertOk (res) {
     throw new Error(message)
   }
```
WeiJun0827, line 18: Not quite sure if the error format of the new endpoint still follows the above flow.

williamchong: https://mainnet-node.like.co/cosmos/tx/v1beta1/txs/0022328B1B61161943373D4C76F386BF8F036C947A30D588243804333F37CF12
just found a random failed tx in explorer, need to modify a bit i guess

williamchong (changes requested):

util/cosmos/misc.js
```diff
@@ -11,14 +11,17 @@ export function assertOk (res) {
...
 
+  const txRes = res.tx_response
   // Sometimes we get back failed transactions, which shows only by them having a `code` property
-  if (res.code) {
-    const message = JSON.parse(res.raw_log).message
+  if (txRes.code) {
+    const message = txRes.raw_log.message
+      ? JSON.parse(txRes.raw_log).message
+      : txRes.raw_log
```
williamchong, line 19: if you can access `txRes.raw_log.message` you don't have to do `JSON.parse`..
if its plaintext, we can't access `.message` until we `JSON.parse`
just use `txRes.raw_log` i guess

williamchong (changes requested):

williamchong (approved):

PR #1620, opened 2022-04-26
✨ Support likeWallet sign-in

nwingt:
This PR should be able to release before chain upgrade

### DIFF
--- a/pages/in/tx/history.vue
+++ b/pages/in/tx/history.vue
@@ -24,8 +24,8 @@
             </div>
             <div class="lc-container-3">
               <view-bigdipper
-                v-if="getUserInfo.cosmosWallet"
-                :address="getUserInfo.cosmosWallet"
+                v-if="getUserWalletAddress"
+                :address="getUserWalletAddress"
               />
               <view-etherscan
                 v-else
@@ -65,8 +65,7 @@ export default {
     };
   },
   asyncData({ store, redirect }) {
-    const user = store.getters.getUserInfo;
-    if (!(user.wallet || user.cosmosWallet)) {
+    if (!store.getters.getUserHasWallet) {
       redirect('/in');
     }
   },
@@ -87,6 +86,7 @@ export default {
       'getUserIsRegistered',
       'getUserHasPendingLike',
       'getUserHasWallet',
+      'getUserWalletAddress',
     ]),
   },
   mounted() {
--- a/pages/in/tx/history.vue
+++ b/pages/in/tx/history.vue
@@ -7,8 +7,8 @@
 
           <transaction-history
             ref="txHistory"
-            :user="getUserInfo.user"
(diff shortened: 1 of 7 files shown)


util/User.js
```diff
@@ -73,12 +73,15 @@ const User = {
     return data;
   },
 
-  async signCosmosLogin(cosmosWallet, signer, platform = 'cosmosWallet', loginMessage = LOGIN_MESSAGE) {
-    if (!cosmosWallet) return null;
+  async signCosmosLogin(inputWallet, signer, platform = 'cosmosWallet', loginMessage = LOGIN_MESSAGE) {
+    if (!inputWallet) return null;
+    if (platform !== 'cosmosWallet' && platform !== 'likeWallet') {
```
williamchong, line 78:
```suggestion
    if (`!['cosmosWallet ', 'likeWallet '].includes(platform)`) {
```

store/modules/getters/user.js
```diff
@@ -12,7 +12,11 @@ export const getUserIsRegistered = state => !!state.user.user;
 
 export const getUserIsInBonusCoolDown = state => state.user.bonusCooldown;
 
-export const getUserHasWallet = state => !!(state.user.wallet || state.user.cosmosWallet);
+export const getUserHasWallet = state => !!(
+  state.user.wallet || state.user.cosmosWallet || state.user.likeWallet
```
williamchong, line 16: Can we remove `state.user.wallet`? This function would be more useful without eth wallet
We need to make sure metamask login flow does not use this function though

williamchong (changes requested):

store/modules/getters/user.js
nwingt: Can we just remove all ETH-related UI except for user login? Otherwise, that will make a lot of change

williamchong: hmm keep this change and leave for revamp?

pages/in/tx/history.vue
```diff
@@ -7,8 +7,8 @@
 
           <transaction-history
             ref="txHistory"
-            :user="getUserInfo.user"
-            :address="getUserHasWallet"
+            :user="getUserId"
+            :address="getUserERC20WalletAddress"
```
williamchong, line 11: The transaction API returns tx of cosmos too, so `getUserWalletAddress || getUserERC20WalletAddress`?
But this UI seems broken anyway

williamchong: Or is this erc20 guard intentional?

nwingt: User without ERC20 address will redirect to /in, I don't think users will come to this page anymore, there is no entry at all in like.co

williamchong:
This PR will go to feature branch, I merge the API in master in a seperate commit i guess

PR #1622, opened 2022-04-28
👽️ Set default to `like` prefix

WeiJun0827:


pages/in/settings/index.vue
```diff
@@ -81,10 +81,10 @@
                     {{ $t('Register.form.walletAddress') }}:
                   </span>
                   <span
-                    v-if="getUserInfo.cosmosWallet"
+                    v-if="getUserInfo.likeWallet"
                     class="content"
                   >
-                    {{ getUserInfo.cosmosWallet }}
+                    {{ getUserInfo.likeWallet }}
```
nwingt, line 87: This is conflict with https://github.com/likecoin/like-co/pull/1620

williamchong: I have merged and rebase the feature branch, @WeiJun0827 please change accordingly thanks

WeiJun0827: Rebased the branch, ready for review, thanks!

nwingt (approved):
I think it is more robust to add cosmosWallet as a fallback instead of replacing them all

williamchong (approved):

williamchong:
We probably want to remove querying cosmosWallet from id/other user object in the future, and only keep the current user part in case user did not migrate

PR #1626, opened 2022-05-03
💄 Update blocking message for chain upgrade

AuroraHuang22:
截圖 2022-05-03 下午5 55 17

### DIFF
--- a/components/toolbars/ToolBars.vue
+++ b/components/toolbars/ToolBars.vue
@@ -97,9 +97,24 @@
         <popup-dialog
           :is-show="isChainUpgrading"
           :allow-close="true"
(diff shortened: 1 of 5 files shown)


williamchong (approved):

components/toolbars/ToolBars.vue
```diff
@@ -97,9 +97,24 @@
...
-        />
+        >
+          <template #footer>
+            <div :style="{ width:'100%',textAlign:'center' }">
+              <i18n
+                path="ChainUpgradingAlertLearnMore"
+                tag="div"
+              >
+                <a
+                  v-t="`this announcement`"
```
williamchong, line 109: what does `this announcement` stand for in this case? would `announcement` do?

PR #1627, opened 2022-05-03
:recycle: check iscn-ar id only like prefix

Tsai-Hsueh-Kuan:
:recycle: Check iscn-ar id only like prefix


williamchong:
https://github.com/likecoin/like-co/pull/1629

PR #1634, opened 2022-06-01
🔧 Lower default gas price

williamchong:


nwingt:
Please rebase to the current master after reverting ytd

PR #1636, opened 2022-06-10
Support WalletConnect in mobile

nwingt:


util/WalletConnect.js
```diff
@@ -119,6 +127,23 @@ class WalletConnect {
     }
   }
 
+  async requestForLogin(loginMessage) {
+    const payload = await Promise.race([
+      this.connector.sendCustomRequest({
+        id: payloadId(),
+        jsonrpc: '2.0',
+        method: 'likerId_login',
```
williamchong, line 135: Is this a special message for iOS to do both init and sign login?

.circleci/config.yml
```diff
@@ -293,6 +292,53 @@ jobs:
...
+            docker tag like-store:latest us.gcr.io/${PROJECT_NAME}/like-store-frontend:legacy-${CIRCLE_SHA1}
+      - run:
+          name: Push images
+          command: |
+            gcloud docker -- push us.gcr.io/${PROJECT_NAME}/like-store-frontend:latest
+            gcloud docker -- push us.gcr.io/${PROJECT_NAME}/like-store-frontend:legacy-${CIRCLE_SHA1}
+      - deploy:
+          name: Deploy
+          command: |
+            kubectl patch deployment like-store-frontend-legacy -p '{"spec":{"template":{"spec":{"containers":[{"name":"like-store-frontend","image":"us.gcr.io/'"$PROJECT_NAME"'/like-store-frontend:legacy-'"$CIRCLE_SHA1"'"}]}}}}'
```
williamchong, line 341: if `legacy` doesn't set any extra build/env var, just the SHA hash would be good?

package.json
```diff
@@ -72,6 +73,7 @@
     "gsap": "^3.0.4",
     "helmet": "^3.21.2",
     "http-proxy-middleware": "^2.0.1",
+    "js-cookie": "^3.0.1",
```
williamchong, line 76: use tiny-cookie too?

williamchong (changes requested):

.circleci/config.yml
nwingt: Do you mean keeping `main` instead of `legacy`?

williamchong: I mean we don't need separated tag for `main` and `legacy`, the commit hash should be enough
on the other hand if we set `IS_TESTNET` or other env in build time, we would need a separated tag

nwingt: Do you want to keep `main` in the tag for mainnet/legacy image?

williamchong: yes please keep `main` vs `test`/`dev`

williamchong (approved):

PR #1637, opened 2022-06-17
Add toggle button for legacy frontend

nwingt:
image

### DIFF
--- a/util/client.js
+++ b/util/client.js
@@ -1,5 +1,5 @@
 import parse from 'url-parse';
-import Cookies from 'js-cookie';
+import { getCookie, setCookie, removeCookie } from 'tiny-cookie';
 import {
   LIKE_BUTTON_POST_MESSAGE_TARGET_ORIGIN,
   REDIRECT_WHITE_LIST,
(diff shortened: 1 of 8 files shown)


util/client.js
```diff
@@ -63,3 +64,14 @@ export function tryPostLoginRedirect({ route }) {
...
 }
+
+const LIKECO_FRONTEND_MODE_COOKIE_KEY = 'likeco_frontend_mode';
+
+export function toggleFrontendMode() {
+  if (Cookies.get(LIKECO_FRONTEND_MODE_COOKIE_KEY)) {
+    Cookies.remove(LIKECO_FRONTEND_MODE_COOKIE_KEY);
+  } else {
+    Cookies.set(LIKECO_FRONTEND_MODE_COOKIE_KEY, 'legacy');
+  }
```
williamchong, line 75: we actually has `vue-cookie` as `this.$cookie`, do you want to use that or just use `js-cookie` ?

nwingt: Perhaps we use [tiny-cookie](https://www.npmjs.com/package/tiny-cookie), which depends on `vue-cookie`

PR #1639, opened 2022-06-21
📈 Add event log for frontend mode toggle

nwingt:


components/dialogs/AuthDialog.vue
```diff
@@ -725,7 +725,10 @@ export default {
       'loginByCosmosWallet',
     ]),
 
-    toggleFrontendMode,
+    toggleFrontendMode() {
+      logTrackerEvent(this, 'RegFlow', 'ToggleFrontendMode', 'ToggleFrontendModeLegacy)', 1);
```
williamchong, line 729: Do we know the exact mode the user is going to here? (`legacy` or `keplr/new`)

williamchong (changes requested):
Can we also set user property on login screen? I am not sure if we are on ua or ga4, but both should have a way to set it
https://developers.google.com/analytics/devguides/collection/ga4/user-properties?technology=websites

components/dialogs/AuthDialog.vue
nwingt: Yes, the new UI is using `AuthDialogV2` component, I will add a tracker in it.

williamchong: so `ToggleFrontendModeLegacy` means from legacy to new?
then `ToggleFrontendModeXXX` will be added meaning from new to legacy?

williamchong (approved):

PR #1655, opened 2022-08-08
💥 Always use like wallet as iscn author id

williamchong:


util/cosmos/iscn/sign.js
```diff
@@ -49,37 +49,35 @@ function getAuthorISCNPayload(payload) {
...
-      entity = {
-        '@id': `${EXTERNAL_URL}/${userId}`,
-        name: displayName,
-      };
-    } else if (likePrefixAddress) {
-      entity = {
-        '@id': likePrefixAddress,
-      };
+    if (userId) {
+      identifier.push({ '@type': 'PropertyValue', propertyID: 'Liker ID', value: `${EXTERNAL_URL}/${userId}` });
```
nwingt, line 61: Is using `liker.land/${user_id}` better for a Depub context?

WeiJun0827 (approved):

util/cosmos/iscn/sign.js
williamchong: this is share across all iscn, so not specific to depub
the question would be should like.co or liker.land be the canonical url for liker id, we assume like.co before

PR #1665, opened 2022-09-22
💄 [Settings] Add popup layout

nwingt:


williamchong (approved):
Does (authcore) user have another page for changing avatar in this case?

nwingt:
> Does (authcore) user have another page for changing avatar in this case?

Yes, it is the same UI
localhost_3000_in_settings_popup

PR #1672, opened 2022-10-10
🐛 Fix legacy metamask login

williamchong:
For some reason web3 method doesn't like what metamask rpc returns now, so I just replace some with direct calls to window.ethereum as suggested in metamask guide


nwingt (approved):
Should merge into legacy branch?

PR #1673, opened 2022-10-11
🚸 Use latest iscn id when updating iscn

williamchong:
To avoid errors like https://www.mintscan.io/likecoin/txs/C681E9B6AA12821BD1EA5E6D8DCC40FFEB85A4859DD5013B361376203E7756C3
In which iscn-ar widget user normally does not care the original version

### DIFF
--- a/util/cosmos/iscn/sign.js
+++ b/util/cosmos/iscn/sign.js
@@ -4,6 +4,7 @@ import { ISCNSigningClient } from '@likecoin/iscn-js';
 import bech32 from 'bech32';
 import { ISCN_RPC_URL } from './constant';
 import { EXTERNAL_URL } from '../../../constant';
+import { getISCNPrefix, getISCNInfoById } from './query';
 
 let isConnected;
 let iscnClient;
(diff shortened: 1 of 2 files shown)


WeiJun0827 (approved):
Seems we have `getISCNPrefix()` in private API, public API, jedi, and like-co now.
May consider refactoring this function into iscn-js, too.

WeiJun0827 (approved):

util/cosmos/iscn/sign.js
```diff
@@ -204,7 +205,9 @@ export async function signISCNTx(
   const client = await getISCNSigningClient(signer);
   let res;
   if (iscnId) {
-    res = await client.updateISCNRecord(address, iscnId, payload, { memo, broadcast });
+    const iscnPrefix = getISCNPrefix(iscnId);
+    const { latestVersion } = await getISCNInfoById(iscnId);
```
nnkken, line 209: AShould use `iscnPrefix` for query instead

williamchong: updated

nnkken (approved):

PR #1680, opened 2022-10-25
👷 Add testnet legacy deployment

williamchong:


.circleci/config.yml
```diff
@@ -195,6 +195,53 @@ jobs:
           name: Deploy
           command: |
             kubectl patch deployment like-api-web-testnet -p '{"spec":{"template":{"spec":{"containers":[{"name":"like-api-web","image":"us.gcr.io/'"$TEST_PROJECT_NAME"'/like-api-public:'"$CIRCLE_SHA1"'"}]}}}}'
+  deploy_rinkeby_legacy_frontend:
+    working_directory: ~/likecoin-store
+    docker:
+      - image: lakoo/node-gcloud-docker
```
nwingt, line 201: Should use our image?

nwingt (approved):

.circleci/config.yml
williamchong: updated

PR #1691, opened 2022-12-08
🚸 Switch back to new frontend after login

nwingt:


williamchong (approved):
Not sure if `toggleFrontendMode` should take an optional param to make sure we are removing cookie

nwingt:
> Not sure if `toggleFrontendMode` should take an optional param to make sure we are removing cookie

@williamchong Updated

williamchong (approved):

PR #1717, opened 2023-06-21
✨ Add likerland as default publisher

williamchong:


util/cosmos/iscn/sign.js
```diff
@@ -135,8 +135,12 @@ export function preformatISCNPayload(payload) {
     recordNotes = '',
   } = payload;
 
-  let stakeholders = inputStakeholders || [];
+  let stakeholders = inputStakeholders ? [...inputStakeholders] : [];
```
WeiJun0827, line 138: If default `inputStakeholders` is `[]`, would `stakeholders` fallback here?

WeiJun0827 (approved):

util/cosmos/iscn/sign.js
williamchong: spreading [] is still [] so it's ok, but spreading undefined is error

WeiJun0827: Since `!!inputStakeholders`  will always be `true`, will `let stakeholders = [...inputStakeholders];` works the same way?

williamchong: The original code is `  let stakeholders = inputStakeholders || [];`
So im not sure if input can be null or false despite having default value

PR #1724, opened 2023-11-23
⬆️ Upgrade node 16, node-sass and stylelint

williamchong:


components/dialogs/ReAuthDialog.vue
```diff
@@ -38,14 +38,14 @@ import {
...
-  },
   components: {
     BaseDialogV2,
     AuthCoreRegister,
   },
+  data() {
+    return {
+    };
+  },
   computed: {
```
nwingt, line 49: what for?

nwingt (approved):

components/dialogs/ReAuthDialog.vue
williamchong: Probably can remove, but lets make another PR...

PR #1741, opened 2024-06-21
🐛 Fix legacy signInPlatform undefined

AuroraHuang22:


pages/in/register/index.vue
```diff
@@ -240,7 +240,19 @@ export default {
...
     async handleAuthSignIn() {
-      const { method, code } = this.$route.query;
+      const { method: initialMethod, code, sign_in_platform: signInPlatform } = this.$route.query;
+      let method = initialMethod;
+      if (!method && signInPlatform) {
+        switch (signInPlatform) {
+          case 'authcore':
+            method = 'liker-id';
+            break;
+          case 'likeWallet':
```
williamchong, line 250: Does this actually exist in legacy code? If not just use default to handle

AuroraHuang22: [🎨 Use likeWallet as default method for legacy platform](https://github.com/likecoin/like-co/pull/1741/commits/4ff6acfa45029cfe6fe9d2c73307f30f248dd5c8)

williamchong (approved):

PR #1746, opened 2024-09-01
🐛 Fix targetRoute name undefined when redirect after sign in

AuroraHuang22:


williamchong:
What was stored if name is not set? A plain string or some other object?

AuroraHuang22:
in most cases, for redirects on liker.land & nft-book-press, it will be `{hash: ‘’}` or null.
This causes the page to loop at the same page after logging in, without redirecting to /settings

nwingt (approved):

PR #1762, opened 2024-10-15
🎨 Use userLikeWallet as loginAddress

AuroraHuang22:


store/modules/wallet.js
```diff
@@ -26,6 +26,8 @@ const state = () => ({
   isInited: false,
   methodType: null,
   isLoggingIn: false,
+
+  accessToken: null,
```
williamchong, line 30:
```suggestion
  accessToken: '',
```

pages/in/widget/pay/index.vue
```diff
@@ -679,21 +679,20 @@ export default {
...
         const amount = new BigNumber(this.totalAmount);
-        const from = await this.fetchCurrentCosmosWallet();
+        const from = await this.fetchCurrentCosmosWallet() || this.getAddress;
         if (!from) {
           throw new Error('PLEASE_RELOGIN');
         }
         if (!this.isUsingKeplr) {
-          const { cosmosWallet, likeWallet } = this.getUserInfo;
-          if ((cosmosWallet || likeWallet) && from !== cosmosWallet && from !== likeWallet) {
+          if (this.getAddress !== this.getUserInfo.likeWallet) {
```
williamchong, line 695:
```suggestion
          if (from !== this.getUserInfo.likeWallet) {
```
Would using `from` here work? Since we defined it above anyway

PR #1801, opened 2025-08-22
[TAS-4388] 👔 Change crisp to intercom

williamchong:


components/footer/Footer.vue
```diff
@@ -140,8 +106,8 @@ export default {
   },
   methods: {
     showCSChat() {
-      if (window.$crisp) {
-        window.$crisp.push(['do', 'chat:open']);
+      if (window.Intercom) {
+        window.Intercom('show');
```
nwingt, line 110: what is the different between this and `$intercom.show()`

